PR #8, opened 2019-06-28
Core project setup

solarissmoke:
Work in progress for setting up a basic wagtail project. Items still remaining:

- [x] Use `django-environ` to manage configuration
- [x] Add release commands for heroku
- [x] Update travis config to run some tests and linting
- [x] Basic documentation for getting started


dockerfiles/Dockerfile.node
```diff
@@ -1,5 +1,12 @@
-FROM node:8-stretch-slim
+FROM node:8-alpine
```
patjouk: Sadly, we have to use Debian and not Alpine because of how VSCode support docker ([note in the install section](https://code.visualstudio.com/docs/remote/containers#_getting-started)). It's still in review on Foundation but should be merged soon.

dockerfiles/Dockerfile.python
```diff
@@ -1,10 +1,20 @@
-FROM python:3.7-slim-stretch
+FROM python:3.7-alpine
```
patjouk: Same here, we need Debian.

dockerfiles/Dockerfile.node
solarissmoke: OK. I switched to alpine because it comes with some of the image processing deps that Wagtail requires. It should be possible to install those manually in the dockerfile though.

patjouk: I think it's safe to remove them: I can run the foundation site on debian without any extra dependencies, both back and front.

donate/manage.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
```
kaedroho: Could this file be moved to the root? This removes the need for ``donate/donate``

solarissmoke: Yep, will do.

README.md
```diff
@@ -1,9 +1,27 @@
...
+
+When it's done, run `docker-compose up`, wait until the static files to be built, and go to `0.0.0.0:8000`. When you want to stop, do `^C` to shut down your containers.
+
+### Running commands inside the Docker container
+
+When the Django server is running, you can start the Django shell with:
+
+    docker-compose exec backend pipenv run python manage.py shell
+
+(TODO: wrap this with invoke to make it less cumbersome).
```
patjouk: I can probably start this today or tomorrow since I know how the project is structured now :)

solarissmoke:
@patjouk this should now provide a working project on heroku. I'd be grateful if you can give it a quick look (also to make sure it works on the various environments you need it to work on for development)?

patjouk:
I will review your PR in details tomorrow, but the review app is up and running, which is a good sign :D

env.default
```diff
@@ -0,0 +1,15 @@
...
+CORS_WHITELIST=*
+CORS_REGEX_WHITELIST=""
+DATABASE_URL=""
+DEBUG=True
+DJANGO_SECRET_KEY=secret
+SET_HSTS=False
+SSL_REDIRECT=False
+USE_S3=False
+X_FRAME_OPTIONS=DENY
+XSS_PROTECTION=True
```
ababic: @solarissmoke Could I suggest adding `REDIS_URL=""` here?

env.default
```diff
@@ -0,0 +1,15 @@
+# Wagtail config
+
+ALLOWED_HOSTS=*
+ASSET_DOMAIN=donate.mofoprod.net
+CONTENT_TYPE_NO_SNIFF=True
+CORS_WHITELIST=*
+CORS_REGEX_WHITELIST=""
+DATABASE_URL=""
```
ababic: Should we add a default here (e.g. `postgres://donate_wagtail:donate_wagtail@localhost:5432/donate_wagtail`)?

If we did, I suppose we'd also have to add `environment` vars for the `postgres` service in `docker-compose.yml`, e.g:

```
postgres:
    image: postgres:11
    ports:
      - "5432"
    environment:
      - POSTGRES_USER=donate_wagtail
      - POSTGRES_PASSWORD=donate_wagtail
      - POSTGRES_DB=donate_wagtail
    volumes:
      - postgres_data:/var/lib/postgresql/data/:delegated

Comment (ababic) [reply env.default]: Or we could keep it simple and stick with the defaults? e.g.
`postgres://postgres:postgres@localhost:5432/postgres)`

Comment (solarissmoke) [reply env.default]: The default currently is to use sqlite in the absence of `DATABASE_URL`, which I think is the most cross-platform compatible option. Mozillians develop on various OS/platform combos and I assumed this is why the default environment defaults to sqlite.

File: env.default (hunk 5 above)
Comment (solarissmoke) [reply env.default]: Could you explain why?

File: env.default (hunk 6 above)
Comment (solarissmoke) [reply env.default]: (note that docker is not the standard development environment, just one option that we support - see http://github.com/mozilla/foundation.mozilla.org/ for the options they use there).

**Comment by patjouk** [comment]
I got swamped today and have to postpone the review to tomorrow :(

File: env.default (hunk 5 above)
Comment (ababic) [reply env.default]: Just to make things a little easier to find. It's quite possible that a developer might want to install and use redis locally to get a more 'production-like' stack. If they did, it's slightly clearer what they need to update in their local `.env` file to make that happen.

File: env.default (hunk 6 above)
Comment (ababic) [reply env.default]: > Mozillians develop on various OS/platform combos and I assumed this is why the default environment defaults to sqlite.

Ah, Okay. sqllite seems like a sensible default in that case :)

File: package.json [hunk 7]
```diff
@@ -0,0 +1,87 @@
...
+    "build:sass": "shx mkdir -p donate/frontend/_css && node-sass source/sass/main.scss donate/frontend/_css/main.compiled.css && npm run autoprefix",
+    "build": "npm run build:common && npm run build:js",
+    "build-uncompressed": "npm run build:common && npm run build:js-uncompressed",
+    "heroku-postbuild": "npm run build",
+    "migrate": "pipenv run python manage.py migrate",
+    "optimize:svg": "find source/images -type f -name '*.svg' -print0 | xargs -0 -n 1 -P 6 svgo --multipass --pretty --enable=removeTitle",
+    "optimize:jpg": "find source/images -type f -name '*.jpg' -print0 | xargs -0 -n 1 -P 6 -I '{}' guetzli --quality 93 '{}' '{}'",
+    "optimize:png": "find source/images -type f -name '*.png' -print0 | xargs -0 -n 1 -P 6 optipng",
+    "optimize": "run-p optimize:**",
+    "percy": "percy exec -t 800 -- npm run cypress",
```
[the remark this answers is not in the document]
alanmoo: FWIW: I do eventually want Percy on donate, but I believe it's outside the scope of the Torchbox work.

docker-compose.yml
```diff
@@ -10,7 +10,7 @@ services:
     environment:
       # Prevents pipenv from loading env files (python dotenv and docker-compose compatibility issue).
       - PIPENV_DONT_LOAD_ENV=1
-    command: echo "I'M A WAGTAIL APP!"
+    command: pipenv run python manage.py runserver 0.0.0.0:8000
```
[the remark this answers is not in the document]
solarissmoke: OK. The current `env.default` works just fine for me though - perhaps because it doesn't have any double quotes in it? I've changed it to look for `.docker.env` anyway.

solarissmoke:
Thanks @patjouk @ababic, I've addressed your feedback.

PR #31, opened 2019-07-10
Fix empty strings in default.env and add missing redis dependency

solarissmoke:
An env var like `REDIS_URL=""` in the env file results in the variable being set as a string `""` (instead of an empty string) - this changes it to not pass any value at all there.

I am not sure if this works across all environments. @patjouk is this something you have dealt with before?


solarissmoke:
As per discussion elsewhere, this is problematic because docker and pipenv (outside docker) treat quotes differently - fixing one breaks the other, so we need to do something [similar to this](https://github.com/mozilla/foundation.mozilla.org/blob/master/tasks.py#L25) for compatibility with both, or stop using pipenv.

patjouk:
I try to reproduce that bug, but apparently, it got fixed: https://github.com/pypa/pipenv/issues/1966
So happy we don't have to do that `.env` dance anymore \o/

PR #34, opened 2019-07-10
We don't need different .env for docker/plain pipenv

patjouk:
Appears that the empty string bug in `.env` files got fixed in [pipenv in april 2018](https://github.com/pypa/pipenv/issues/1966) \o/ Pipenv and docker-compose can now share the same `.env`.


solarissmoke:
Could you update the [instructions here](https://github.com/mozilla/donate-wagtail/blob/master/README.md#how-to-setup-your-dev-environment-with-docker) as well?

patjouk:
@solarissmoke Done!

PR #36, opened 2019-07-11
Landing page - front-end

nicklee:
Based on feature/landing-page, initial FE work for the landing page, including basic grid setup


source/sass/components/_button.scss
```diff
@@ -7,6 +7,17 @@
     &:hover {
         cursor: pointer;
     }
+
+    &--primary {
+        background-color: $color--primary;
+        color: $color--white;
+        transition: background-color $transition;
+
+        &:hover {
```
chris-lawton: does this need `:focus` & `:active` as well or are they defined elsewhere?

source/sass/components/_link.scss
```diff
@@ -1,8 +1,12 @@
 .link {
-    color: $color--link;
 
-    &:hover {
-        color: $color--link-hover;
+    &--light-bg {
+        color: $color--link;
+
+        &:hover {
```
chris-lawton: same Q as above

source/sass/components/_button.scss
nicklee: Needed here, added

PR #33, opened 2019-07-10
Add Landing Page model and scaffolding for populating data to review apps

solarissmoke:
Closes #20. See also #11.

Notes for review:

1. Wireframes for this page: https://redpen.io/qmb0c00a7348799ffe

2. I've put the landing page in a `core` app because I think we are likely to reuse a lot of the functional view logic that will be added here for the campaign pages as well, and so want them to live in the same place.

3. We've used the same approach as on [foundation.mozilla.org](https://github.com/mozilla/foundation.mozilla.org/) for producing fake data for review apps - management command run on a `postdeploy` hook (but haven't added the Slack integration that that project has at this stage).

4. This doesn't include the form, which I think will be best rendered manually in the template, and will be done as part of #22.

Issue #19, opened 2019-07-10
Site-wide data models


solarissmoke:
I propose to use site-wide settings to store the following:

- Legal footer text: rich text (required)
- Top navigation menu links: orderable sequence of link blocks that can link either to a page on site or to an absolute URL (required)
- Footer links: orderable sequence of link blocks that can link either to a page on site or to an absolute URL (required)
- Language code to currency mapping (as defined in #1).
- Currency information:
    - Code
    - Symbol
    - Allowed payment methods
    - Default donation amounts: for each currency, 4 default suggestion values for that currency.

It might be easier to store currency information in code instead - to make a call before starting work on this.

solarissmoke:
I just realised that putting the legal footer text into a setting will complicate translations. If this text is not likely to change often then the simplest approach in my view is to hard-code it into the template, so that it can be translated in the normal way.

alanmoo:
The top nav is only in English because the Foundation site itself is mainly only in English.

solarissmoke:
I've been looking into how to store currency information, and feel that we're best off having a hard-coded dictionary very similar to [the current one](https://github.com/mozilla/donate.mozilla.org/blob/master/src/data/currencies.js). 

The rationale for this is that this data doesn't change frequently, and so it doesn't seem worth the cost of modeling in the database and making editable in Wagtail. 

@alanmoo @TheoChevalier does this sound OK to you?

TheoChevalier:
@solarissmoke No issue on my side, thanks for asking :)

solarissmoke:
Closing on the basis that we've ended up storing all of this information in code, so no site-wide models necessary at present.

alanmoo:
Out of curiosity (because we might need it on our other Wagtail site) is there any documentation on best practices for site-wide model implementation?

solarissmoke:
Yes - use `wagtail.contrib.settings`: https://docs.wagtail.io/en/v2.5.1/reference/contrib/settings.html

PR #37, opened 2019-07-11
Add Campaign Page model and initial currency data.

### LINKED ISSUE #21: Campaign page model
### Field spec

- Hero image: image field (required)
- Title: plain text (required)
- Lead text: plain text (required)
- ~Partner logos: orderable sequence of image fields (optional)~
- Default donation amount overrides: optional list of items mapping a currency to list of 4 suggested donation amounts. 

### Notes

- If a currency is not listed in the donation amount overrides, then we will fall back to the [global defaults](https://github.com/mozilla/donate-wagtail/issues/19).
- Legal text is fetched from global configuration (i.e., does not vary per-campaign).

solarissmoke:
Depends on #33. See #21.

This adds the model fields and basic template for campaign pages. It also adds basic currency data [taken directly from the current site](https://github.com/mozilla/donate.mozilla.org/blob/master/src/data/currencies.js) - opted to hard-code this rather than model it because it doesn't change frequently (see discussion on #19).

The campaign page can override currency defaults, and models for this are included here. Implementation of the logic to use these will happen when we work on the braintree forms.


donate/core/models.py
```diff
@@ -23,3 +28,69 @@ class LandingPage(Page):
...
+    # Define 4 fields each for single and monthly preset amounts
+    preset_kwargs = {'max_digits': 7, 'decimal_places': 2}
+    single_1 = models.DecimalField(**preset_kwargs)
+    single_2 = models.DecimalField(**preset_kwargs)
+    single_3 = models.DecimalField(**preset_kwargs)
+    single_4 = models.DecimalField(**preset_kwargs)
+    monthly_1 = models.DecimalField(**preset_kwargs)
+    monthly_2 = models.DecimalField(**preset_kwargs)
+    monthly_3 = models.DecimalField(**preset_kwargs)
+    monthly_4 = models.DecimalField(**preset_kwargs)
```
ababic: @solarissmoke Any thoughts about using a couple of simple `StreamField`s here instead, or are we happy always having exactly 4 single/monthly options each time?

alanmoo: Interesting question. I could see us experimenting with more/fewer fields in the future; Is there any more/less significant overhead using a streamfield? (Sure, slight variations in database, UX of creating a campaign, but probably not major?)

ababic: The only clear downside I can think of is that you'd have to manually click an 'Add' button a few times in order to enter the number of values you need (I'm not sure it's possibly to display 4 items on load). The additional computational cost of parsing the json string into a list of decimal values would be negligible.

It would make the values a little trickier to access directly in view/test code I supposed, but it would cut out a lot of duplication here, and in the template where they are output.

alanmoo: Mmmm yes, testing. As long as we can still do that robustly, I'd be for the streamfield. (Obviously the CSS would have to be a bit forgiving, but that's nothing a little flexbox can't fix.)

solarissmoke: A streamfield would work from a data perspective - I had actually implemented this before reverting to a hard coded list. 

The reason I did that was because all the defaults are hard-coded, 4 each (and there is a large number of currencies). This means that if you wanted to change to say 5 options, you would either have to do it for *all* currencies, or allow different numbers of options for different currencies, and a proper way of handling showing/hiding extra options in JS on the front end. This is all doable but adds complexity to that logic.

What do you think @alanmoo ? Open to going either route.

alanmoo: Ah, great point. That makes me lean towards leaving it as-is, but I’ll confirm with Will tomorrow (he ended up needing today off too).

alanmoo: Will was very interested in the Streamfield route, so let's go with that. I'm happy to hop on a call to hash out the details on this if needed (or at least understand what decisions need to be made) as early as tomorrow at 9AM EDT, just send me an invite.

solarissmoke: As per discussion yesterday, we're going with a streamfield.

jafacakes2011:
LGTM

PR #42, opened 2019-07-11
add env vars for using s3 on review apps

patjouk:
Review apps inherit their s3 config for staging.


app.json
```diff
@@ -20,11 +20,26 @@
...
+    },
+    "AWS_S3_CUSTOM_DOMAIN": {
+      "required": true
+    },
+    "AWS_SECRET_ACCESS_KEY": {
+      "required": true
+    },
+    "AWS_STORAGE_BUCKET_NAME": {
+      "required": true
+},
```
solarissmoke: Small indentation error here.

PR #49, opened 2019-07-12
Add flow for collecting single card payment with Braintree

solarissmoke:
This PR adds:

1. All the basic scaffolding required for Braintree integration, including a base mixin that all flows will derive from, and that does some core error handling.
2. A working flow for collecting a single card payment - starting from `/pay/`. This includes a step for collecting user address information and a simple thank you page.

Work to be done separately:

1. Link landing/campaign pages to the payment flow introduced here.
2. Add support for multiple currencies (everything is currently in USD).
3. Other secondary components such as upsell to monthly, email signup.


donate/payments/views.py
```diff
@@ -0,0 +1,246 @@
...
+    template_name = 'payment/personal_details.html'
+
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
+
+    def get(self, request, *args, **kwargs):
+        # Ensure that the donation amount is a valid currency decimal
+        # (either only a whole number, or a number with up to 2 decimal places)
+        if not re.match(r'^\d+(?:\.\d{1,2})?$', self.request.GET.get('amount', '50')):
+            return HttpResponseRedirect(self.campaign_source_page.get_url(request))
```
solarissmoke: Note to self - `campaign_source_page` isn't implemented yet and I need to remove this until it is.

donate/payments/views.py
```diff
@@ -0,0 +1,246 @@
...
+    def get_success_url(self):
+        return reverse('payments:card_single')
+
+
+class SingleCardPaymentView(BraintreePaymentMixin):
+    template_name = 'payment/card_single_payment.html'
+
+    def dispatch(self, *args, **kwargs):
+        self.personal_details = self.request.session.get('personal_details')
+        # If we don't have donation_personal_details in the session, the user
```
ababic: minor typo here

donate/payments/views.py
```diff
@@ -0,0 +1,246 @@
...
+        })
+        return ctx
+
+
+class PersonalDetailsView(FormView):
+    form_class = PersonalDetailsForm
+    template_name = 'payment/personal_details.html'
+
+    def dispatch(self, request, *args, **kwargs):
+        return super().dispatch(request, *args, **kwargs)
```
ababic: I think we can probably kill this override :)

donate/payments/views.py
```diff
@@ -0,0 +1,246 @@
...
+
+from . import gateway
+from .exceptions import InvalidAddress
+from .forms import BraintreePaymentForm, PersonalDetailsForm
+from .utils import freeze_personal_details_for_session
+
+logger = logging.getLogger(__name__)
+
+
+class BraintreePaymentMixin(FormView):
```
ababic: If this is subclassing a full-fledged view class like `FormView`, I would suggest going with `BraintreePaymentView` as a name here instead. A 'mixin' should be more agnostic about the other views it can be combined with.

source/js/payments.js
```diff
@@ -0,0 +1,116 @@
...
+            }
+        );
+    }
+
+    initHostedFields();
+}
+
+document.addEventListener('DOMContentLoaded', function() {
+    setupBraintree();
+});
```
ababic: Not quite sure how to review this :) Is there a javascript styleguide anywhere that we're supposed to be adhering to? It could be a good idea to have a front-end dev cast an eye over this.

solarissmoke: @chris-lawton would you be able to review this JS please? We have ESLint set up on the project but it looks like it has to be run separately - I will do that next week.

source/js/payments.js
```diff
@@ -0,0 +1,116 @@
...
+        errorDiv.toggleAttribute('hidden', true);
+        errorDiv.innerHTML = '';
+    }
+
+    function showFieldError(container) {
+        container.classList.add('braintree-hosted-fields-invalid');
+    }
+
+    function clearFieldError(container) {
+        container.classList.remove('braintree-hosted-fields-invalid');
```
chris-lawton: could assign `'braintree-hosted-fields-invalid'` to a var

source/js/payments.js
chris-lawton: Looks good @solarissmoke!

solarissmoke: Thanks! There are linting issues in this JS and in the styles, which I will make a separate PR for. Also need to add linting to travis.

PR #60, opened 2019-07-12
Add invoke tasks

patjouk:
I added a bunch of invoke tasks to make our life easier. The only downside is that you have to install invoke system-wide :(

The node tests were failing: it looks like it's missing a config file. When it's fixed, we will need to uncomment the `inv docker-test-node` tasks.

edit: I've just discovered GitHub hub and tried to use it for the first time: sorry for that weird title XD


patjouk:
I can also remove `settings_test.py` for now if you want to have this merged.

PR #83, opened 2019-07-22
Update JS dependencies

patjouk:
I did a bunch of updates for our node dependencies. Still waiting on a fix from chokidar-cli (lodash issue).


patjouk:
Thanks for the review :)

PR #88, opened 2019-07-23
[WIP] Documentation for Basket

patjouk:
I've added some documentation on how to talk to Basket.
@solarissmoke do you need more details?

A bit of historical context: the current setup for the newsletter signup is a bit more complicated than just sending a POST request to basket API. When someone signup, a JSON message is sent to a different queue than the one for donations (it's due to an old limitation of [a tool we used](https://github.com/jbuck/hatchet)). Then, [sawmill](https://github.com/mozilla/sawmill/blob/master/worker/send_post_request.js), a queue processor, get the message and sent a POST request to the basket newsletter API. As you might have seen in Basket donations processor, there's also a an event type for signing up people to a newsletter: it's used for the foundation site only for now. There are ongoing discussions on how this should evolve and @alanmoo is meeting with @pmac to discuss this.

CC @stephaniemcv

close MozillaFoundation/mofo-devops#694


patjouk:
I'm changing this ticket to WIP because @alanmoo mentionned that some changes might be coming to basket.

solarissmoke:
This looks OK to me. Braintree exposes the data used to determin `conversion_amount`, `net_amount` and `transaction_fee` in a slightly different way to stripe - but I think we should be able to manipulate it into this structure.

README.md
```diff
@@ -69,3 +75,78 @@ When the Django server is running, you can start the Django shell with:
...
     docker-compose exec backend pipenv run python manage.py test --settings=donate.settings_test
+
+
+## Basket
+
+[Basket](https://github.com/mozmeao/basket) is a tool run by MoCo to manage newsletter subscriptions and donations. It's listening for messages (JSON) sent to a SQS queue.
+
+### Basket donations queue:
+
+Basket has 4 [donation event types](https://github.com/mozmeao/basket/blob/master/basket/news/management/commands/process_donations_queue.py#L21-L26):
```
alanmoo: Are these more like "4 event types the foundation uses"?

README.md
```diff
@@ -69,3 +75,78 @@ When the Django server is running, you can start the Django shell with:
...
+- `donation`: process a donation and send its data to SFDC,
+- `crm_petition_data`: add petition signature to SFDC,
+- `newsletter_signup_data`: newsletter signup for the foundation site,
+- `DEFAULT`: process a followup Stripe event on a donation.
+
+For this project, we're only using the `donation` event type.
+
+### Donation event type
+
+Example of a donation message sent to Basket:
```
alanmoo:
```suggestion
Example of a donation message sent to Basket, via SQS as that's known to work right now and a donation endpoint isn't yet built into Basket:
```

README.md
```diff
@@ -69,3 +75,78 @@ When the Django server is running, you can start the Django shell with:
...
+`donation_url`: url from which the donation was made,
+`locale`: language code for the donor,
+`conversion_amount`: donation amount in USD, before transaction fees,
+`net_amount`: donation amount in USD, after transaction fees,
+`transaction_fee`: payment processor's transaction fees in USD
+
+
+### Newsletter signup
+
+We're using Basket newsletter API to signup people to our newsletter. Specs are available in [Basket's documentation](https://basket.readthedocs.io/newsletter_api.html#news-subscribe)
```
alanmoo:
```suggestion
We're using Basket newsletter HTTP API to signup people to our newsletter. Specs are available in [Basket's documentation](https://basket.readthedocs.io/newsletter_api.html#news-subscribe). While we could use the SQS, the Basket team has expressed interest in moving towards standard HTTP requests for all Basket interaction. (A donation event endpoint will be added in the future.)
```

README.md
patjouk: I would go for "Basket has 4 event types we can use".

alanmoo:
```suggestion
Basket has [4 event types we can use](https://github.com/mozmeao/basket/blob/master/basket/news/management/commands/process_donations_queue.py#L21-L26): 
```

README.md
patjouk: I wouldn't add that: Pmac was very elusive when I ask them if they plan to rewrite that component. But let's add more content about SQS and how that work.

README.md
patjouk: Same as the previous comment: let's not take about possible implementation in the future.

README.md
alanmoo:
```suggestion
Example of a donation message sent to Basket, via SQS:
```

README.md
alanmoo:
```suggestion
We're using Basket newsletter HTTP API to signup people to our newsletter. Specs are available in [Basket's documentation](https://basket.readthedocs.io/newsletter_api.html#news-subscribe). (Note that this is different than the SQS approach used for donation events) 
```

README.md
```diff
@@ -69,3 +75,78 @@ When the Django server is running, you can start the Django shell with:
...
+  created: 1563801762,
+  recurring: false,
+  service: 'stripe',
+  transaction_id: 'ch_1Ez1TSG8Mmx3htnxyShib70n',
+  project: 'mozillafoundation',
+  last_4: '4242',
+  donation_url: 'http://localhost:3000/en-US/',
+  locale: 'en-US',
+  conversion_amount: 50,
+  net_amount: 48.6,
```
solarissmoke: Looking at what information Braintree [provides for a transaction](https://developers.braintreepayments.com/reference/response/transaction/python), I don't think we will be able to determine the transaction fee (we get the Paypal transaction fee, but not Braintree's own fee for card payments) and therefore the net amount.

alanmoo:
I’ve reached out to Braintree about net amount. @solarissmoke

README.md
pmac: 🕵️‍♂️

alanmoo:
@patjouk If you accept my suggestions, are we good to land this?

solarissmoke:
Hey @patjouk @alanmoo , would be great if we could finalise this soon, and then set up a staging basket instance that we can send some data to?

pmac:
If you want to post some test data [basket-dev](https://basket-dev.allizom.org) is a good option since it is connected to a sandbox Salesforce instance.

alanmoo:
I believe this is all set; @pmac does it look right to you?

@solarissmoke i think you’re good for a Basket staging environment, do you have the SQS staging environment you need too?

pmac:
Yeah. Looks good to me 👍

solarissmoke:
@alanmoo I don't think we have an SQS environment yet - grateful if you can arrange that.

solarissmoke:
@alanmoo were you able to talk to Braintree about transaction fees etc? I still don't think it will be possible to determine `net_amount` and `transaction_fee`.

alanmoo:
I think you're right, and I think that's something that we're going to punt on. The CRM should be used for CRM purposes anyway, not for accounting.

Issue #13, opened 2019-07-09
Detail page


nicklee:
**Design / Wireframe:**

Desktop: https://redpen.io/jc07fc459ff5f3fe81
Tablet: https://redpen.io/gpa264abeea7f4c799
Mobile: https://redpen.io/ykdc92900c06af901c

**Components that form this template:**

- [x] Detail form
- [x] Legal text

nicklee:
Designs:

Desktop: https://redpen.io/skc2914b634e141270
Tablet: https://redpen.io/gpa264abeea7f4c799
Mobile: https://redpen.io/pk3fa92bc99a0ddc32

nicklee:
MR here https://github.com/mozilla/donate-wagtail/pull/90

beccaklam:
Sorry @philwakefield not sure exactly what you need from me here?

philwakefield:
What should the user see and do when they click 'change amount'. This is currently un-styled in above link and assumes the right functionality.

beccaklam:
Example of change amounts dialogue: https://redpen.io/ju45b724a9d5bd3cfe
<img width="337" alt="Screen Shot 2019-07-25 at 4 16 01 PM" src="https://user-images.githubusercontent.com/20977376/61905645-8ac78f00-aef7-11e9-81d9-62d68c55a63f.png">

nicklee:
Thanks! We've just implemented this, shall post a staging link once it's all merged in.

nicklee:
This is all done, only outstanding item i'm aware of is outputting a "per month" value to the change amount area when the user is doing a monthly donation. For example "You are donating $ 17 per month", is this something you're able to do @solarissmoke ?

solarissmoke:
Yep - see https://github.com/mozilla/donate-wagtail/pull/133

PR #90, opened 2019-07-23
Feature/13 detail page fe

nicklee:
Styling for the detail page: 

https://github.com/mozilla/donate-wagtail/issues/13

Everything done other than change amount styling, which is waiting for a design.

Also contains branch `feature/14-header-footer-fe`


donate/templates/pages/base_page.html
```diff
@@ -47,7 +47,8 @@
             </div>
 
             <a class="header__logo-link" href="/">
-                <img class="header__logo" src="/_images/mozilla-logo-white.svg" alt="Mozilla">
+                <img class="header__logo header__logo--mobile" src="/_images/mozilla-logo-m.svg" alt="Mozilla">
+                <img class="header__logo header__logo--desktop" src="/_images/mozilla-logo-white.svg" alt="Mozilla">
```
chris-lawton: should these not be referenced using the `{% static 'path/to/image' %}` tag or has it been setup differently in this codebase?

source/js/components/donation-amount-toggle.js
```diff
@@ -0,0 +1,30 @@
+class AmountToggle {
+  static selector() {
+    return "[data-amount-toggle]";
+  }
+
+  constructor() {
+    this.toggleButton = document.querySelector("[data-amount-toggle]");
```
chris-lawton: rather than re-selecting `[data-amount-toggle]` you can just pass `node` into the constructor and assign it to `this.toggleButton`, e.g.

```
constructor(node) {
    this.toggleButton = node;
```

source/js/main.js
```diff
@@ -29,9 +30,13 @@ document.addEventListener("DOMContentLoaded", function() {
...
+    AmountToggle.selector()
+  )) {
+    new AmountToggle(donatetoggle);
+  }
+
   for (const tabs of document.querySelectorAll(Tabs.selector())) {
     new Tabs(tabs);
   }
-
-  tabIndexer();
```
chris-lawton: was this deleted on purpose as it looks like the function still exists? Or is it being called elsewhere?

donate/templates/pages/base_page.html
solarissmoke: It's set up a bit differently here because of how we're using whitenoise.

nicklee: We've updated these in the static tooling branch, will merge in

source/js/main.js
nicklee: Good spot, shouldn't have been removed, thanks!

PR #97, opened 2019-07-25
Add support for upgrading single card payments to monthly subscriptions.

solarissmoke:
Work in progress for card upsell. The following is pending/awaiting clarification:

- [x] Whether upsell is only available for some currencies, and if so how to determine those.
- [x] Applying an initial amount to the form.


donate/payments/utils.py
```diff
@@ -12,3 +13,16 @@ def freeze_transaction_details_for_session(details):
...
+
+def get_suggested_monthly_upgrade(currency, single_amount):
+    info = get_currency_info(currency)
+
+    # Check if we have upgrades manually specified for this currency, and if so use that
+    for tier in info.get('monthlyUpgrade', []):
+        if Decimal(single_amount) >= tier['min']:
+            return Decimal(tier['value'])
+
+    # If we don't have manual values, or they didn't return a suggestion, then
```
solarissmoke: @alanmoo let me know if this sounds OK to you. Use `monthlyUpgrade` where set, otherwise fall back to 10% of the single amount.

alanmoo: Yes, but we need a minimum amount, which is partially why we don’t ask until we hit a threshold. Otherwise there’s potential for recurring small donations to cost us money.  (We also need that for the custom amount field. If that wasn’t built in, I’m happy to file it)

solarissmoke: I'm merging this to unblock Nick on templates, and moving this discussion to the feature ticket so that we can make the appropriate adjusments before closing that.

donate/templates/payment/card_upsell.html
```diff
@@ -0,0 +1,53 @@
...
+                            {% endfor %}
+                        {% elif form.errors %}
+                            <div class="form__errors">
+                                {% trans "There were some errors with your form. Please amend the fields highlighted below." %}
+                            </div>
+                        {% endif %}
+
+                        {% render_form_field form.amount %}
+
+                        <p>{% trans "I authorize Mozilla Foundation to charge my card in the amount shown below and to continue doing so each month until I cancel" %}
```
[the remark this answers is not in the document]
solarissmoke: Indeed it should, thanks.

PR #99, opened 2019-07-25
Feature/donate form tabs

nicklee:
- Updating tab js to work with radio buttons instead of anchors
 - Adding in custom select box styling


source/sass/components/_form-custom-select.scss
```diff
@@ -0,0 +1,69 @@
...
+    display: block;
+    width: $select-height;
+    height: $select-height;
+    pointer-events: none;
+    background-color: #fff;
+  }
+
+  select {
+    @include font-size(xs);
+    -webkit-appearance: none;
```
chris-lawton: do you have autoprefixer? I thought this would have been generated by using `appearance: none;` which would give you the `-moz-` prefix as well as `-webkit-`. Either way, you'll need the `-moz-` one in there :)

nicklee: Thanks, this was me bug bashing, and forgot to remove (the actual select is hidden by the custom one, so the appearence doens't really matter :))! We do have autofixer, i'll get it removed

chris-lawton:
Looks good @nicklee :)

PR #101, opened 2019-07-25
Feature/landing page responsive fe

nicklee:


nicklee:
Would be great if you could review my srcset, can't work out if it's a caching thing or just a browser limitation. Srcset works fine as you resize down, but if you resize back up it seems to retain the mobile version of the image. @chris-lawton

Issue #1, opened 2019-06-26
Confirm approach for localisation and currency detection


solarissmoke:
A few queries about localisation and currency defaults:

1. Is browser language preference used to determine the default language presented to the user?

2. How should the default currency be determined? Is this currently determined from location (e.g., using IP geolocation)?

3. How should we determine suggested donation amounts in each currency? Will these vary on a per-campaign basis?

alanmoo:
> 1. Is browser language preference used to determine the default language presented to the user?

I believe so. Check out the current implementation [here](https://github.com/mozilla/donate.mozilla.org/blob/master/src/lib/get-locale.js), which is called by [location-parser.js](https://github.com/mozilla/donate.mozilla.org/blob/master/src/lib/location-parser.js) @theochevalier do you have anything useful to add here? (One gotcha: if you see a reference to `/public/locales.json`, that file is generated at build time, so you might need to spin up the old repo locally to see what it looks like if needed.)

> 2. How should the default currency be determined? Is this currently determined from location (e.g., using IP geolocation)?

We currently rely on locale (as a pairing of language and region), and then read what currency we should use based on that, from [locale-data.js](https://github.com/mozilla/donate.mozilla.org/blob/master/src/data/locale-data.js). This approach seems to work well, so I can't imagine we'd want to change it.

> 3. How should we determine suggested donation amounts in each currency? Will these vary on a per-campaign basis?

Right now we can pass in [query strings](https://github.com/mozilla/donate.mozilla.org/blob/master/docs/Query_Strings.md) if we want to show specific amounts, but by default we pull from [currencies.js](https://github.com/mozilla/donate.mozilla.org/blob/master/src/data/currencies.js) for everything. (This file may be auto generated based on exchange rates; I can dig into that if you'd like). That being said, @WillatMozFdn may have specifics on the ideal spot for controlling these amounts (query params, campaign basis, etc.). My instinct is as long as we have as much flexibility as we do with query params, the approach isn't too important (though I know there's a certain amount of usefulness in being able to simply write a URL and know it will do the intended thing because of the query params).

alanmoo:
I just had a follow-up conversation with @WillatMozFdn regarding item three. The query string functionality, as it stands today, is necessary to replicate. (There's some additional params that would be nice, but I believe are outside the scope of this work, and the internal team can add things like that in the future). 

Two other  that we identified would be useful, but not necessary:
[...]

PR #111, opened 2019-07-26
Add currency detection functionality.

solarissmoke:
Add currency detection functionality to set the initial currency as [defined here](https://github.com/mozilla/donate-wagtail/issues/1#issuecomment-509908160):

1. If a currency is specified in URL query string, use that.
2. Otherwise, if a currency is mapped to user's exact language code (e.g., en-ca, use that).
3. If not, look for a mapping for the base language (e.g., es if the code is es-es), and if one exists use that.
4. If all else fails, default to USD.


solarissmoke:
Thanks for highlighting that! I _think_ the desired behaviour is to derive a currency from each valid language, and have added a test for that. @alanmoo please let me know if this makes sense to you?

So for example, if the accept-language header is:

    fo-FO;q=0.9, ko;q=0.8

(where `fo-FO` is some language that we don't recognise, and `ko` is Korean) then we should default to Korean Won because that is the first language we recognise (as opposed to defaulting to USD after we failed to recognise the first language).

solarissmoke:
I'm going to merge this with the behaviour described above so that Nick is able to finish up the JS side of things - please leave a note here if the behaviour needs to change and I'll address that in a new PR.

PR #116, opened 2019-07-26
Feature/donation amount toggle fe

nicklee:
JS to toggle the change amount form, then if it's submitted to update donation amounts on the page (in final submit button, and in notification at the top of page). Also added a hidden input that's populated with the value, so BE can process it.


donate/templates/payment/card.html
```diff
@@ -15,24 +15,26 @@
...
-                            <input type="number" name="donation-update" value="{{ form.amount.value }}" step="1" min="1">
+                    <form class="donation-update__container">
+                        <div class="donation-update__form-item form-item">
+                            <label class="donation-update__label form-item__label" for="donation-update">{% trans "Donation Amount" %}</label>
+                            <div class="donation-update__currency-symbol">{{ currency_info.symbol }}</div>
+                            <input class="donation-update__input form-item__input" id="js-update-donation-value" type="number" name="donation-update" value="{{ form.amount.value }}" step="1" min="1">
                         </div>
                         <div class="form-item">
-                            <button class="button button--primary">{% trans 'Confirm' %}</button>
+                            <button class="donation-update__action button button--primary button--large" id="js-update-button">{% trans 'Confirm' %}</button>
```
solarissmoke: Could we update `id_amount` (rendered in line 66 of this file) instead?

source/js/components/donation-amount-toggle.js
```diff
@@ -5,14 +5,33 @@ class AmountToggle {
...
+    this.hiddenInput = document.querySelector("#js-donation-value");
     this.activeClass = "active";
+    this.selectedClass = "selected";
 
     this.bindEvents();
   }
 
+  UpdateHiddenInputFromInput() {
+    let input = document.querySelector("#js-update-donation-value");
+    this.hiddenInput.value = input.value;
```
solarissmoke: From experience on sams I think we need to check that this input is valid - i.e., that it can be parsed as a number - before we update the value.

donate/templates/payment/card.html
nicklee: Done

source/js/components/donation-amount-toggle.js
nicklee: I've added some validation

donate/templates/payment/card.html
```diff
@@ -15,24 +15,25 @@
...
                 <div class="donation-update" data-amount-actions>
-                    <div class="donation-update__container">
-                        <div class="form-item">
-                            <label for="donation-update">{{ currency_info.symbol }}</label>
-                            <input type="number" name="donation-update" value="{{ form.amount.value }}" step="1" min="1">
+                    <form class="donation-update__container">
+                        <div class="donation-update__form-item form-item">
+                            <label class="donation-update__label form-item__label" for="donation-update">{% trans "Donation Amount" %}</label>
+                            <div class="donation-update__currency-symbol">{{ currency_info.symbol }}</div>
+                            <input class="donation-update__input form-item__input" id="js-update-donation-value" type="number" name="donation-update" value="{{ form.amount.value }}" step="1" min="1">
```
solarissmoke: I think the `step` attribute will be problematic as it prevents setting decimal amounts, which they need the ability to do - e.g., $3.14 on Pi day. Maybe we just set `min` and round the value to 2 decimal places in JS?

nicklee: Seems like step isn't an issue, this is simple how many to go up when the number input's arrows are pressed. The user can still enter 3.14 if they want, I've added some js to ensure the number can only be up to 2 decimal places though.

Issue #102, opened 2019-07-25
Other amount donation selection


nicklee:
The other amount field will require some js to select the checkbox when focused, and also to populate a hidden input with the amount entered.

nicklee:
Done in #118

nicklee:
Example can now be viewed on staging:
https://donate-wagtail.mofostaging.net/en-us/card/single/?currency=usd&amount=50

PR #118, opened 2019-07-29
Feature/currency forms

nicklee:
JS to populate donation form


donate/templates/fragments/donate_form.html
```diff
@@ -150,7 +110,6 @@ <h2 class="donate-form__actions-heading heading heading--secondary">{% trans "Ch
         {% render_form_field braintree_form.braintree_nonce %}
         {% render_form_field braintree_form.amount %}
         {% render_form_field braintree_form.frequency %}
-        {% render_form_field braintree_form.currency %}
```
solarissmoke: This is a hidden input which is required by the Paypal JS - see https://github.com/mozilla/donate-wagtail/blob/master/source/js/payments-paypal.js#L8 . Without it the Paypal modal will error when you try to complete payment. Was this deleted by mistake?

donate/templates/fragments/donate_form.html
```diff
@@ -19,30 +19,10 @@
...
-                        <label for="one-time-amount-4" class="donation-amount__label">$3</label>
-                    </div>
-                    <div class="donation-amount donation-amount--two-col donation-amount--other">
-                        <input type="radio" class="donation-amount__radio" name="amount" value="other" id="one-time-amount-other" autocomplete="off">
-                        <label for="one-time-amount-other" class="donation-amount__label">$</label>
-                        <input type="text" class="donation-amount__input" id="one-time-amount-other-input" placeholder="{% trans 'Other amount' %}">
-                    </div>
-                    <input type="hidden" class="one-time-amount__total">
+                <div class="donate-form__fields" id="js-donate-form-single">
+                    {# Filled with js #}
```
solarissmoke: Do you think we should render the initial set of options in the back end, and then only use JS for subsequent changes? I am not sure it matters a great deal in the sense that nothing will work anyway if JS is disabled, bit it would result in a faster render for initial page load.

source/js/components/currency-selector.js
```diff
@@ -0,0 +1,115 @@
...
+          type === "monthly-amount" ? "per month" : ""
+        }
+                    </label>
+                </div>`;
+      })
+      .join("");
+
+    container.insertAdjacentHTML(
+      "beforeend",
+      `<div class='donation-amount donation-amount--two-col donation-amount--other'><input type='radio' class='donation-amount__radio' name='amount' value='other' id='${type}-other' autocomplete='off'><label for='${type}-other' class='donation-amount__label' data-currency>$</label><input type='text' class='donation-amount__input' id='${type}-other-input' placeholder='Other amount'></div>`
```
solarissmoke: This value is not currently being passed to the next step - let's discuss today now to achieve this?

solarissmoke: Oh, just noticed the separate ticket for that in #102

nicklee: yea, seperated this up, hoping to look at this today, I think we'll also need to handle the decimal place stuff in there too

donate/templates/fragments/donate_form.html
nicklee: I ended up adding a min-height to the form (assuming the form has 4 radio options plus other field) to ensure there was no initial jump on load. But it might make sense to render this via BE if it's easy to do. Like you say, would increase the initial page load!

donate/templates/fragments/donate_form.html
nicklee: Yep, looks like a mistake (no idea how that happened!) thanks for spotting it, have re-added.

source/js/components/currency-selector.js
```diff
@@ -0,0 +1,115 @@
...
+    return "#id_currency-switcher-currency";
+  }
+
+  constructor(node) {
+    this.selectMenu = node;
+    this.formContainer = document.getElementById("js-donate-form");
+    this.data = JSON.parse(document.getElementById("currencies").innerHTML);
+    this.oneOffContainer = document.getElementById("js-donate-form-single");
+    this.monthlyContainer = document.getElementById("js-donate-form-monthly");
+    this.defaultCurrency = document.getElementById(
```
solarissmoke: @nicklee I reckon that just `document.getElementById("id_currency-switcher-currency").value` does the trick here.

nicklee: Updated

PR #124, opened 2019-07-30
Thank you, Email newsletter and donation upsell FE

nicklee:


beccaklam:
@nicklee Build is looking awesome! 😎 Just some small nit picky design things:

## P1
1. For tablet width (768 and above?) there should be at least 48px padding or margin on both sides of the content
2. The line-height is too large for mobile: The header line height should be around 1.1 with at 16px padding or margin underneath. For body copy (html) can we do a line height of 1.4 instead of 1.6 for mobile? Also don't know what solution is best so that the code is not too finicky. Open to suggestions.
Screen Shot 2019-07-30 at 11 10 46 AM

## P2
1. This text should be Zilla Slab Light (font-weight 300)
Screen Shot 2019-07-30 at 10 55 25 AM

PR #134, opened 2019-07-31
Keep track of which currencies AMEX is not available for.

solarissmoke:
This alters the structure of the `disabled` key in currency data to a list - we'll need to tweak the currency switcher JS to handle this.


nicklee:
Added new js to allow for multiple

PR #135, opened 2019-07-31
Add render amount options for initial currency in the template.

solarissmoke:


donate/templates/fragments/donate_form.html
```diff
@@ -20,7 +20,20 @@
...
+                    {% for amount in initial_currency_info.presets.single %}
+                    <div class="donation-amount">
+                        <input type="radio" class="donation-amount__radio" name="amount" value="{{ amount }}" id="one-time-amount-{{ forloop.counter0 }}" autocomplete="off">
+                        <label for="one-time-amount-{{ forloop.counter0 }}" class="donation-amount__label">
+                            {{ initial_currency_info.symbol }}{{ amount }}
+                        </label>
+                    </div>
+                    {% endfor %}
+
+                    <div class="donation-amount donation-amount--two-col donation-amount--other">
```
solarissmoke: @nicklee could you please check that this is rendered correctly? There are some empty data attributes which perhaps shouldn't be. Also now that we are rendering in the template, it may be possible to reduce what is [rendered here](https://github.com/mozilla/donate-wagtail/blob/master/source/js/components/currency-selector.js#L78) to only modify the existing HTML instead of replacing it?

nicklee: Looks good to me. Think we're best keeping this as is for the time being, maybe we can refactor if we have time at the end? The custom value js all working now, and is related to this (we also use .innerHTML to replace all the content, so this would need to be rejigged too)

source/js/components/currency-selector.js
```diff
@@ -140,9 +134,6 @@ class CurrencySelect {
     this.selectMenu.addEventListener("change", () => {
       this.getSelectValue();
     });
-
-    // Initial defaults
-    this.processSelectDefaultValue();
```
nicklee: Have added this back in, we need to check if the default value has any payment methods disabled, so I've moved this into this call.

PR #141, opened 2019-08-01
Update currency handler to pass selected currency to card forms.

solarissmoke:
@nicklee please could you review this?

I have noticed an issue now that no amount option is selected by default - if you try to submit the card button you will get silently redirected back to the page, with no indication why. Paypal will also fail - I think we either need to default to one of the options, or add validation that checks if an option is set before we allow submission of either button. What do you think is the best approach?


nicklee:
The designs show an item selected, with this in mind, I've added the code to make the first item selected by default.

Everything else looks good to me!

Issue #30, opened 2019-07-10
Post-payment email sign up


solarissmoke:
Provide the option for the user to sign up to the newsletter after they have completed payment.

### Information required

1. Where should the form data be sent (SQS)?

alanmoo:
@patjouk Can you share the SQS we use for Basket on donate prod/staging presently? (I'm not sure if that should be shared via github or not)

patjouk:
I'm adding more details on #27 but yes, that data goes to SQS too, to be processed by basket: https://github.com/mozmeao/basket/blob/master/basket/news/management/commands/process_donations_queue.py#L24 and https://github.com/mozilla/donate.mozilla.org/blob/master/src/components/signup.js#L38-L41

solarissmoke:
This looks straightforward enough - thanks.

patjouk:
good news: this is actually documented in basket docs: https://basket.readthedocs.io/newsletter_api.html#news-subscribe

Here is an example of a message we're currently sending:
```
      { format: 'html',
        lang: 'en-US',
        newsletters: 'mozilla-foundation',
        trigger_welcome: 'N',
        source_url: 'https://donate.mozilla.org/',
        email: 'bob@bob.org',
        country: undefined }
```

solarissmoke:
@alanmoo I think it was mentioned briefly somewhere, but I can't find a record of it: I believe you need the ability to skip the newsletter step (e.g., if someone is already signed up to the newsletter). Was this to be done by setting a query argument on the landing page - and is there a particular argument you want us to use?

alanmoo:
Correct. If they arrive on the donate site with ‘subscribed=1’ don’t ask for newsletter signups.

solarissmoke:
Closing on the basis that newsletter signups are now in place - let's open new issues for specific data structure issues after testing.

PR #142, opened 2019-08-01
Add ability to skip newsletter signup by setting a query arg on landing page

solarissmoke:
As described here: https://github.com/mozilla/donate-wagtail/issues/30#issuecomment-516849584

### DIFF
--- a/donate/core/tests/test_models.py
+++ b/donate/core/tests/test_models.py
@@ -24,6 +24,16 @@ class DonationPageTestCase(TestCase):
             'clp'
         )
 
+    def test_serve_sets_subscribed_cookie(self):
+        request = RequestFactory().get('/?subscribed=1')
+        response = DonationPage().serve(request)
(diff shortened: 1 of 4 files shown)


donate/core/tests/test_models.py
```diff
@@ -24,6 +24,11 @@ def test_get_initial_currency_uses_header(self):
             'clp'
         )
 
+    def test_serve_sets_subscribed_cookie(self):
+        request = RequestFactory().get('/?subscribed=1')
+        response = DonationPage().serve(request)
+        self.assertEqual(response.cookies['subscribed'].value, '1')
+
```
ababic: @solarissmoke Since the setting of the cookie is conditional, could we also add a test to show that it is NOT set when 'subscribed' is not present, or has a value other than '1'?

solarissmoke: Yep, good point. Added.

Issue #140, opened 2019-07-31
Currency amounts internationalization


TheoChevalier:
Not sure if this is already scoped/planned as part of another ticket, so filing it just to make sure.

Amounts + currency strings everywhere on the website will need to be formatted using `Intl.NumberFormat` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat

This API will handle the currency symbol formatting, but also convert numbers to the appropriate script and pick separators (decimal, thousands, etc).

Example in Arabic, where numbers are converted to the Arabic script
![image](https://user-images.githubusercontent.com/1294206/62238860-b81aae00-b3cb-11e9-8c51-4dbfc1b9d697.png)

I think the current website uses a npm package, but that package relies on this API ultimately (which is definitely the way to go)

This API would also solve #93 (commented in there with the relevant example + method)

TheoChevalier:
@solarissmoke Thanks for the PR! I had a look on staging, the JS part seem to work great, but I think there are issues with moneyed (or I didn’t test correctly)

With this query for instance, paying in KRW from the French page, 
https://donate-wagtail.mofostaging.net/fr/card/single/?source_page_id=3&currency=krw&amount=22320
I’m getting `₩22,320` while I should see `22 320 ₩` (which is what you get when accessing https://donate-wagtail.mofostaging.net/fr/ and switching currency to KRW)

Side question, what’s moneyed’s source of data? It seem pretty much homemade, no?
If that helps, Babel seem to provide similar features, and be regularly updated to include latest CLDR data http://babel.pocoo.org/en/latest/api/numbers.html

TheoChevalier:
Ah, just noticed you filed #163 a while ago, this seem to be the issue I’m hitting

solarissmoke:
Hey @TheoChevalier 

> Ah, just noticed you filed #163 a while ago, this seem to be the issue I’m hitting

No, that's a different issue to do with overriding the suggested donation _amounts_. The point you have raised is separate. I think you're quite right that babel is a better choice here than moneyed, and will be very easy to swap out. I think that will solve the issue and will let you know once done.

solarissmoke:
I think this is all working as expected now, so closing this issue - so that we can track any edge cases on new tickets.

Issue #93, opened 2019-07-24
More flexible handling of currency symbols


solarissmoke:
From discussion on #82:

> A lot of locales will want the currency symbol to be displayed after the value, would it be possible to add a conditional for that? Maybe storing that preference in settings?
> 
> To give you an idea: https://transvision.mozfr.org/string/?entity=mozilla_org/mozorg/home/index-quantum.lang:3fe118b0ed61665a878f4a81ef2d3357&repo=mozilla_org

TheoChevalier:
Actually, using this would probably be a much cleaner and more reliable way of handling this issue https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat/formatToParts#Examples

Looking for the array index of the currency symbol should let you know where to display it in the UI.

If this works, then we would not need to store and maintain that information

solarissmoke:
I will look into this today. I think we also need to render this stuff correctly in the templates, and am looking at [moneyed](https://github.com/limist/py-moneyed/) to help us do that. In combination, that plus the number formatting API in JS should provide everything we need.

solarissmoke:
This should now all be working - let's address any edge cases in separate tickets.

PR #149, opened 2019-08-02
Add locale-based currency formatting.

solarissmoke:
See #140, #93. This PR adds locale-based currency formatting in two ways:

1. Currencies rendered in Django templates are formatted using `moneyed`.

2. Values manipulated in JS are formatted using `Intl.NumberFormat`.

Worth noting that `Intl` is not supported in IE10 and below - if this is a problem we're going to need a polyfill.

I haven't removed the currency symbols defined in `constants.py` - ideally we would rely on `moneyed` to provide these, but it doesn't make this at all easy to do outside of a locale context, so I've gone with using our symbols as a fallback.

@nicklee I have a small display issue, where the space between "Donate " and the currency is being lost on the button on the card details view - this I _think_ is due to flex styling but I am not sure how best to resolve. Grateful if you can have a look at that and also review the JS changes.

### DIFF
--- a/source/js/components/currency-selector.js
+++ b/source/js/components/currency-selector.js
@@ -6,6 +6,9 @@ class CurrencySelect {
   constructor(node) {
     this.selectMenu = node;
     this.formContainer = document.getElementById("js-donate-form");
+    // Convert a locale provided by Django in the form en_US to a
+    // BCP 47 compliant tag in the form en-US.
+    // See https://tools.ietf.org/html/bcp47
     this.locale = this.formContainer
       .getAttribute("data-locale")
       .replace("_", "-");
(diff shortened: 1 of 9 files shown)


solarissmoke:
@TheoChevalier I'd appreciate it if you can review this approach?

nicklee:
@solarissmoke I've just pushed a fix for the flex spacing issue, if you can check it works at your end that would be great.

solarissmoke:
Yep, works now, thank you!

source/js/components/currency-selector.js
```diff
@@ -6,6 +6,9 @@ class CurrencySelect {
   constructor(node) {
     this.selectMenu = node;
     this.formContainer = document.getElementById("js-donate-form");
+    this.locale = this.formContainer
+      .getAttribute("data-locale")
+      .replace("_", "-");
```
ababic: @solarissmoke I would suggest adding a note here to clarify why this transformation is necessary

TheoChevalier:
> @TheoChevalier I'd appreciate it if you can review this approach?

Sorry, kind of post-mortem review :) 
I’m going to file a follow-up in the issue as this PR has been merged

PR #151, opened 2019-08-02
Add worker for processing tasks to send data to Basket.

solarissmoke:
The WIP specification for data being sent to basket is here: https://github.com/mozilla/donate-wagtail/pull/88 - this is based on that.


solarissmoke:
@patjouk are you OK for us to add the Redis-to-go app on Heroku?

patjouk:
Yes!

donate/payments/views.py
```diff
@@ -534,6 +551,15 @@ def get(self, request, *args, **kwargs):
             return HttpResponseRedirect(self.get_success_url())
         return super().get(request, *args, **kwargs)
 
+    def form_valid(self, form):
```
ababic: @solarissmoke How do you feel about adding a `add_newsletter_subscription_to_basket=True` as a keyword argument to this method? It would only be used by tests currently, but using it would be much simpler all the additional mocking, and would be pretty self-explanatory (less need for comments throughout).

solarissmoke: Yes, this is a good idea, thanks!

PR #171, opened 2019-08-07
Drop py-moneyed in favour of babel.

solarissmoke:
As [suggested by @TheoChevalier here](https://github.com/mozilla/donate-wagtail/issues/140#issuecomment-518725204), this PR replaces py-moneyed with the more robust and well-maintained babel library for currency formatting.

Note that babel has a limitation, which is that it isn't possible to tell its `format_currency` utility to ignore decimals for integer values (unless we provide custom formats for all currencies), so I have had to do some manipulation to achieve that. I _think_ this will work for all currencies, but it is possible that there are edge cases that we may run into.

@TheoChevalier would be grateful for your review on this.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -23,7 +23,7 @@ boto3 = "*"
 braintree = "*"
 django-countries = "*"
 freezegun = "*"
-py-moneyed = "*"
+babel = "*"
 
 [requires]
 python_version = "3.7"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "dc67954be6c1395a597117d87054e7b29427657c55b1f5bd234393c382ec29ba"
+            "sha256": "76f11589b0639bae4f3401b89ef4b460e44aebfd745cfc9a732a3f3625578a2e"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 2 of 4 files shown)


TheoChevalier:
Thanks @solarissmoke! Looks pretty good, I’ve only noticed one tiny issue with the value in the button for this query https://donate-wagtail-staging-pr-171.herokuapp.com/fr/card/single/?source_page_id=3&currency=krw&amount=22320

For some weird reason, only in Firefox (tried with a release build as well), and only for the button (no issue with the “You are donating 22 320 ₩” at the top), the space separator is not displayed. But you can see it’s there. And it works in Chrome 🤷‍♂️

Capture d’écran 2019-08-07 à 10 49 06

TheoChevalier:
The space here may or may not be a non-breaking space, if that helps debugging

solarissmoke:
The space displays correctly for me on Firefox (68) on Ubuntu - what OS are you on?

TheoChevalier:
> The space displays correctly for me on Firefox (68) on Ubuntu - what OS are you on?

Right, that seems to be an issue specific to macOS. I can reproduce on two different macOS machines (10.14.6), but can’t reproduce on a Windows 10 one, nor on a Ubuntu VM…

solarissmoke:
I'm going to merge this functional change, and make a separate issue for the display issue that you've noticed on macOS.

PR #185, opened 2019-08-15
Additions to data sent to basket and Braintree custom fields

solarissmoke:
This PR adds logic to pass some additional data to Basket and Braintree custom fields:

- Landing page URL and project identifier to basket
- Campaign ID and project identifier to Braintree

I ended up passing this data through the original forms, even though campaign ID and project ID could be obtained from the source page. The reason for this is that this information is required, and so we don't want to rely on the source page being available after the user has started a payment flow.

@nicklee FYI I have removed the `name` attribute from the `one-time-amount-other-input`, because if present this pollutes the GET params on the next step. Lighthouse will complain but I think we'll just have to ignore that.

### DIFF
--- a/donate/templates/fragments/donate_form.html
+++ b/donate/templates/fragments/donate_form.html
@@ -1,4 +1,4 @@
-{% load i18n form_tags util_tags %}
+{% load i18n util_tags %}
 <div class="donate-form__content" id="js-donate-form" data-locale="{% get_locale %}">
     <div class="donate-form__tabs tabs js-tabs">
         <form class="donate-form__tab-options tabs__container" role="tablist">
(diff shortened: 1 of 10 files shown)


donate/templates/fragments/donate_form.html
```diff
@@ -142,6 +142,9 @@ <h2 class="donate-form__actions-heading heading heading--secondary">{% trans "Ch
         {% render_form_field braintree_form.frequency %}
         {% render_form_field braintree_form.currency %}
         {% render_form_field braintree_form.source_page_id %}
+        {% render_form_field braintree_form.landing_url %}
+        {% render_form_field braintree_form.project %}
+        {% render_form_field braintree_form.campaign_id %}
```
ababic: @solarissmoke Would it be much more work to add/use a `render_hidden_form_fields` tag that output a hidden input for everything in `form.hidden_fields`?

solarissmoke: No, it wouldn't, and this has the additional benefit that the field would get rendered without the useless template wrapper that it currently gets. I'll amend to do this, thanks!

PR #188, opened 2019-08-16
Refactor currency detection logic to use only user locale.

### LINKED ISSUE #166: Direct access to a localized page should override the detected browser locale
For instance, using a French browser, if I access directly to https://donate-wagtail.mofostaging.net/pl/, the page locale (and associated currency) should ignore the browser locale and switch to Polish instead (and propose Złotys).

Of course, if I was accessing https://donate-wagtail.mofostaging.net/ instead, I should get redirected to https://donate-wagtail.mofostaging.net/fr/ and get EUR by default (this currently works as expected)

solarissmoke:
Closes #166 - refactor currency detection to rely only on the locale determined by Django (and not the 'Accept-Language` header).

cc @TheoChevalier

### DIFF
--- a/donate/core/models.py
+++ b/donate/core/models.py
@@ -50,8 +50,8 @@ class DonationPage(Page):
         # Query argument takes first preference
         if request.GET.get('currency') in constants.CURRENCIES:
             return request.GET['currency']
-        # Otherwise sniff based on browser language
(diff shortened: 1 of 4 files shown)


TheoChevalier:
The new behavior looks good, thanks!

PR #203, opened 2019-08-21
SQS config for basket

patjouk:
- add env vars in `env.default",
- add env vars to review apps.

I added the env vars to staging too.
We can now test sending messages to SQS and basket.


env.default
```diff
@@ -52,3 +52,9 @@ BASKET_SQS_QUEUE_URL=
 RECAPTCHA_SITE_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
 RECAPTCHA_SECRET_KEY=6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
 RECAPTCHA_ENABLED=False
+
+# SQS for Basket
+BASKET_QUEUE_REGION=
```
solarissmoke: We don't currently use this environment variable in the code base, on the basis that the URL should include the region? 

The queue URL is currently fetched from `BASKET_SQS_QUEUE_URL` (not `BASKET_QUEUE_URL`) - see line 48 above. We'll need to change one or the other. We also need to specify a root URL for the API (line 47 above).

We also don't currently use either of `BASKET_ACCESS_KEY` and `BASKET_SECRET_ACCESS_KEY` - can you confirm what these are for? From what I could tell the newsletter signup API doesn't require authentication?

patjouk:
Managed to completely miss that we already had an env var for basket SQS >.<

Upon Samir advice, I created a new AWS user with all the permissions with we need for s3 and SQS. Staging env vars are already updated.

solarissmoke:
@patjouk thanks for this! I've tested on the review app, and it looks like data is being submitted to SQS (transactions) and to the basket API (newsletter signups). Would you be able to check that this data is being processed correctly by Basket?

patjouk:
@pmac Can you check on basket side how the tests went? Look for a test donation with  'campaign_id': 'Section artist writer simple. Such number several concern structure gas.' (12:18 UTC) and a newsletter subscription from "test@example.com" (12:24 UTC).
Thanks.

pmac:
to which SQS queue and basket instance were these tests submitted?

solarissmoke:
@pmac the newsletter signups are going to https://basket-dev.allizom.org, and the SQS queue is https://sqs.us-east-1.amazonaws.com/152582112771/donate-mozilla-org-sfdc-queue-staging

pmac:
I don't see anything in salesforce for `test@example.com` but they may just be rejecting the invalid domain. I'd recommend testing with real addresses or testing addresses at restmail.net (an email testing service Mozilla runs). Also I can't look up records by Campaign ID. What email was used for that petition signature?

solarissmoke:
I'm going to merge this so that this is available to test on staging - I suspect there will be further tweaks required once someone is able to inspect what is actually being sent to SQS.

PR #209, opened 2019-08-23
Refactor URL prefix rewriting

### LINKED ISSUE #207: "Live" links from Wagtail route to the wrong language prefix
Links to view live versions of a page from the CMS result in a 404, because Wagtail generates a URL in the form `/en-us/` instead of `/en-US/`.

solarissmoke:
Fixes #207.

The approach I had taken originally to get custom URL prefixes to work turned out to have issue - specifically that while we can control middleware and the i18n patterns, we can't override Django's behaviour when reversing URLs. 

The way this was addressed on [bedrock](https://github.com/mozilla/bedrock/blob/bf67e8bf342935b4c2e20662ef1ed9a135624aa2/bedrock/base/urlresolvers.py) was to write custom utilities for all of this - the problem with that is that third party apps (e.g., Wagtail) have no awareness of them. I don't think that solution is viable here.

So instead I've done with modifying two core Django utilities in place. This considerably simplifies how much work we have to do - with the caveat that this sort of modification comes with its risks. I can't see another way around this though, other than getting the fundamental issue fixed upstream in Django.


solarissmoke:
Merging in order to get staging working again - this will still need review at some point.

Issue #206, opened 2019-08-22
FAQ pages


alanmoo:
We need to preserve our [FAQ page](https://donate.mozilla.org/en-US/faq). They don't need to be stored in Wagtail models, but they do need to be hosted on the new repo. [Source from old repo](https://github.com/mozilla/donate.mozilla.org/blob/master/src/pages/faq.js)

solarissmoke:
@alanmoo are you OK with us creating a standard content page (translatable) in the CMS, that can be used for this and anything else in future? It will take just as long as hard-coding this particular template, and leave us more flexibility for future.

alanmoo:
That sounds reasonable; At first glance it looks like there's a small amount of logic in that page around when we show different questions, but I'm not entirely sure it's necessary to preserve it all. (Maybe it's just around the env key and captcha, which is fine logic to just ditch and always show the question.)

solarissmoke:
Hmm, having looked at the page, it has quite specific styling and JS. I'll talk to @nicklee to discuss how best to implement.

solarissmoke:
There's also this: https://donate.mozilla.org/en-US/ways-to-give#check . Think it's best we make a content type for it.

nicklee:
@alanmoo do you know where are these pages going to be linked from? Do we need to add them to the footer?

TheoChevalier:
> @alanmoo do you know where are these pages are going to be linked from? Do we need to add them to the footer?

They are already properly linked here (+ in email receipts)
![Screenshot_2019-10-08 Donate now](https://user-images.githubusercontent.com/1294206/66384617-fb3e6200-e9b6-11e9-947e-ee9dc412a0ce.png)

nicklee:
Great thanks @TheoChevalier, I've just merged the FAQ styled in https://github.com/mozilla/donate-wagtail/pull/424

alanmoo:
Blocked by #469

PR #210, opened 2019-08-23
Add ContentPage model and template, for use with general content pages.

solarissmoke:
Refs #206. This adds a general `ContentPage` model that can be used for arbitrary content (see https://donate.mozilla.org/en-GB/faq, https://donate.mozilla.org/en-GB/ways-to-give).

I'll have to ask @nicklee to add some styling to this.


alanmoo:
Is this going to require us to localize the entire paragraphs at once, rather than smaller strings?

solarissmoke:
> Is this going to require us to localize the entire paragraphs at once, rather than smaller strings?

Bump to @kaedroho

PR #333, opened 2019-10-01
install gettext tools (localization fix)

patjouk:
Let's try to copy msgfmt first and see if it's enough.


TheoChevalier:
If this works, you should be able to see “ONE TIME” and “MONTHLY” translated on /fr

bin/post_compile
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+GETTEXT_SHA256="2f991a49b3fb616a032490c907cb700966c5fd998d2aa12b86d547a93746ed66"
+GETTEXT_TARBALL="https://assets.mofoprod.net/gettext/gettext-0.20.1.tar.xz"
```
cadecairos: FYI this was compiled in a Heroku Dyno and uploaded to the mofo-assets s3 bucket for us to use here.

bin/post_compile
```diff
@@ -0,0 +1,33 @@
...
+curl -s -L -o gettext.tar.xz $GETTEXT_TARBALL
+
+# Verify the contents of the package
+if [ "$GETTEXT_SHA256" != "$(sha256sum gettext.tar.xz | cut -d ' ' -f 1)" ]; then
+    echo "Checksum doesn't match. gettext tools will not be installed."
+    exit 1
+fi
+
+# Extract the gettext tools
+tar -xvf gettext.tar.xz > /dev/null
```
Pomax: no need for `-v` if we're sending the output to dev/null

Issue #272, opened 2019-09-25
Enable support for Braintree's advanced fraud tools


alanmoo:
We need AFT on for our account: https://developers.braintreepayments.com/guides/advanced-fraud-tools/configuration

philwakefield:
@solarissmoke please confirm status of this. Not clear on ticket

solarissmoke:
@philwakefield awaiting code review.

solarissmoke:
This is now in place. You can test on sandbox using the [test card number provided here](https://developers.braintreepayments.com/guides/advanced-fraud-tools/testing-go-live/python) - which should result in a failed payment.

There are some important notes to be aware of [about testing in production](https://developers.braintreepayments.com/guides/advanced-fraud-tools/testing-go-live/python#test-transactions-in-production).

PR #346, opened 2019-10-02
Add advanced fraud detection support for card payments.

solarissmoke:
Refs #272. 

AFT must be enabled on the Braintree dashboard before this change is merged - otherwise the card payment form will fail to load properly. Also worth noting that:

- This check is only done on the initial payment form, and not on the upsell form - by that stage we have already done a fraud check on the card, and it doesn't seem like there is any value in repeating it again.

- It's not done on Paypal payments, as the fraud tools are not compatible with that method [according to the Braintree documentation](https://developers.braintreepayments.com/guides/advanced-fraud-tools/overview#compatibility). I guess Paypal does its own checks.


cadecairos:
@alanmoo want me to enable advanced fraud detection support on our sandbox and production braintree accounts?

alanmoo:
As soon as this is approved, yes, go ahead and take over merging/enabling @cadecairos

alanmoo:
Looks like there's a conflict here @solarissmoke

solarissmoke:
@alanmoo please can you enable AVS on the sandbox account, for us to verify that this works on the review app.

solarissmoke:
@alanmoo / @cadecairos this is good to go when you are - just need to enable AFT in the sandbox, then test that the review app works before we merge.

alanmoo:
Will pointed out to me that our Braintree transactions don't seem to indicate device data is being captured (meaning AFT aren't being passed to Braintree, though @cadecairos observed it does seem like we're running the code client-side.) Any thoughts on why this isn't doing the right thing?

image

solarissmoke:
Can't see why this would happen - [we are passing the device data to Braintree](https://github.com/mozilla/donate-wagtail/blob/ea33505539b1e0d97f2a0249bd084cc91deee4ad/donate/payments/views.py#L192). I'll have a look tomorrow to inspect what exactly is being sent - we might need to ask Braintree support to help understand why that isn't being processed.

solarissmoke:
@alanmoo turns out we weren't actually sending device data for single transactions (only for monthly). I have fixed that in #517, but even after doing so I still get "Device Data Captured: False" for all transactions. I have verified that the data is being sent to Braintree:

```
'device:data': '{"device_session_id":"6654d1bd32d9d66905b057fefd5ccaa3","fraud_merchant_id":"600000"}'
```

Think it might be worth asking them if there's something we're missing? Also worth noting that the missing `device_data` should have failed the transaction entirely according to their documentation, so it's possible that things on sandbox don't work the same as in production.

alanmoo:
I've reached out to Braintree, as I made a real transaction in production after deploying this and it still does the same thing.

Issue #328, opened 2019-10-01
Fix contrast ratio for upsell form <input>


youriwims:
### Issue: [Elements must have sufficient color contrast](https://dequeuniversity.com/rules/axe/3.2/color-contrast?application=AxeChrome)

**Element path:** `.form-item__input-additional-info`

**Snippet:** `<span class="form-item__input-additional-info">/ Monthly</span>`
<img width="277" alt="Screen Shot 2019-09-30 at 9 14 12 PM" src="https://user-images.githubusercontent.com/36484608/65927416-a18ed400-e3c7-11e9-8bfb-1032b3c527f6.png">

**How to fix:**
Fix any of the following:
-  Foreground and background colors do not currently meet the minimum WCAG 2 AA contrast ratio thresholds.

====

More Info:

Comments: Updating text `color` to `#666666` fixes the issue.
Resources: n/a

nicklee:
Updated to #666

youriwims:
Perfect 👍

PR #359, opened 2019-10-02
Updating monthly upsell note from #999 to #666

nicklee:


nicklee:
https://github.com/mozilla/donate-wagtail/issues/328

Issue #316, opened 2019-09-30
Revise semantic structure for header elements


youriwims:
### Issue: [Banner landmark must not be contained in another landmark](https://dequeuniversity.com/rules/axe/3.2/landmark-banner-is-top-level?application=AxeChrome)
**Element path:** `.header`,`.newsletter-signup__header`,`.footer-links__header`

**Snippet:** _multiple instances_

**How to fix:**
>By default, the HTML5 `<header>` element has an identical meaning to the banner landmark, unless it is a descendant of `<aside>`, `<article>`, `<main>`, `<nav>`, or `<section>`. [MDN]

Fix the following:
-  reorganize semantic structure to make non-banner `<header>` elements child of one of the above tags
- ensure banner `<header>` element is not inside another banner element
```
<header class="app__header">
    <header class="header">
```

====

More Info:

Comments: n/a
Resources: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_role

nicklee:
I've updated the .header to a div, retaining app__header as the header.

youriwims:
There are also 2 headers in the footer `.newsletter-signup__header` & `.footer-links__header` can we fix those as well? An easy fix would be to update their parent `<div>` to one of the following:  `<aside>` or `<section>`.

nicklee:
oops, missed those, shall get this updated

youriwims:
Cool, thanks!

nicklee:
I've updated the two header tags to divs, hope that's OK

youriwims:
Yep, that works as well. 👍

youriwims:
Everything looks good on staging.

PR #365, opened 2019-10-02
Updating header to div

nicklee:
https://github.com/mozilla/donate-wagtail/issues/316


youriwims:
@nicklee https://github.com/mozilla/donate-wagtail/issues/316#issuecomment-537547115

Issue #297, opened 2019-09-30
Gracefully handle incorrect query params


cadecairos:
I was able to trigger a 5xx error by manually changing the amount value in the URL bar: https://donate-wagtail-staging.herokuapp.com/en-CA/card/single/?source_page_id=3&currency=cad&amount=99999999999999999999999999999999999999

Let's try to handle that gracefully, i.e. overriding it to a predetermined max value (on that note, I shouldn't be able to donate a billion dollars: https://donate-wagtail-staging.herokuapp.com/en-CA/card/single/?source_page_id=3&currency=cad&amount=1000000000)

cadecairos:
This also seems to be an issue with the `source_page_id` param.

philwakefield:
Point 1 - agree
Point 2 - we do not recommend setting a max as this would have to be set for every currency.

cadecairos:
On our old donate platform we configured max amounts in our Form/Stripe/backend to help reduce fraud. This may no longer be necessary with Braintree, so lets not worry about that now and get those other error cases handled.

solarissmoke:
We've added some validation both client- and server-side to guard against silly amounts, and added logic to avoid exceptions if they do somehow get through (#380). The source_page_id issue is fixed in (#353).

cadecairos:
looking good on staging! thanks @solarissmoke!

PR #380, opened 2019-10-03
Add some validation to prevent ridiculously large amounts on payment forms.

solarissmoke:
Refs #297. 

@nicklee I've made somechanges to handle cases of stupidly large amounts that were triggering 500 errors. One of these involves changing the "other amount" field to a number input (which I think there is a separate task for already) - but this has the side effect of introducing up/down toggle buttons on the input, which we might not want. 

I assume there is a way to style these away? Do we first need to confirm that they are not wanted?


nicklee:
Hi @solarissmoke, I've added some styling to hide the number spinner, so it looks the same as previously.

PR #396, opened 2019-10-03
Legacy stripe webhooks

cadecairos:
~Still working on a testing strategy, since sending test webhooks from Stripe fails because it uses fake Stripe Object IDs.~

Tests are done.

[Original webhook code for reference](https://github.com/mozilla/donate.mozilla.org/blob/8a428fde8dd25eef0eb177fe9ff08b17d2abff59/server/routes/index.js#L868-L971)


Pipfile
```diff
@@ -32,6 +32,7 @@ django-rq = "*"
 redlock = {git = "https://github.com/glasslion/redlock.git",ref = "7f873cc362eefa7f7adee8d4913e64f87c1fd1c9"}
 django-csp = "*"
 sentry-sdk = "==0.13.2"
+stripe = "*"
```
Pomax: Is this a safe version?

donate/payments/tasks.py
```diff
@@ -131,6 +169,99 @@ def process_dispute_lost(self, notification):
         })
 
 
+class StripeWebhookProcessor:
+    def process(self, event):
+        # Stripe event types use dot-notation, so convert periods to underscores
+        event_type = event.type.replace('.', '_')
+
+        process_method = getattr(self, 'process_{}'.format(event_type), None)
```
Pomax: `f'process_{event_type}` rather than format? (given the use of formatted strings below, too)

donate/payments/tasks.py
```diff
@@ -131,6 +169,100 @@ def process_dispute_lost(self, notification):
...
 
+class StripeWebhookProcessor:
+    def process(self, event):
+        # Stripe event types use dot-notation, so convert periods to underscores
+        event_type = event.type.replace('.', '_')
+
+        try:
+            process_method = getattr(self, f'process_{event_type}')
+        except AttributeError:
+            logger.info(f'An unsupported Stripe event was not processed: {event_type}')
```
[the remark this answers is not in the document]
cadecairos: I don't think so that's necessary. This leaves us room to add/remove event types without generating errors.

PR #398, opened 2019-10-04
Add support for selecting a different merchant account for Paypal based on donation amount

### LINKED ISSUE #211: Ensure correct Paypal account is used based on donation amount
Apparently it's a well known workaround for Paypal's fees that it's useful to have two different accounts- one for smaller transactions, and one for larger ones. We previously implemented this [here](https://github.com/mozilla/donate.mozilla.org/blob/master/server/lib/paypal-account-switcher.js) and should do the same in this project.

solarissmoke:
Fixes #211.

@alanmoo there are a couple of things to note here:

1. I have copied the logic from [the current site](https://github.com/mozilla/donate.mozilla.org/blob/9823bb7a671ede8a267e850faf00ff0f767db601/server/lib/paypal-account-switcher.js) to determine which account to use. The basic logic here is:

      - Assume micro payments come with a transaction fee of 6%, plus a fixed fee which is currency specific.
      - Assume macro payments come with a transaction fee of 2.9%, plus a fixed fee which is currency specific.

    Paypal's [published fees](https://www.paypal.com/us/webapps/mpp/merchant-fees#fixed-fees-standard) are slightly more complex in that they vary depending on whether the payment is coming from within the US. The micro transaction fee is either 5% or 6.5% depending on this.

    I'm assuming that the current calculations are basically approximating an average of these, but let me know if you want to refine further. I think it is also worth someone going through and checking that the fixed fees for each currency (defined in `payments/constants.py`) are still correct.

2. I can't find any information from Paypal about charges for subscriptions - so have assumed that the same rates apply, and on that basis subscriptions are created using the micro/macro account using the same logic as above, based on the amount of the monthly subscription.

To test this:

1. On the review app, define a new env var `BRAINTREE_MERCHANT_ACCOUNTS_PAYPAL_MICRO`, which is in the same format as `BRAINTREE_MERCHANT_ACCOUNTS`. It specifies which merchant account to use for each currency, for micro transactions.

2. Make small ($1) and large ($20) payments, and verify that these are put through the correct Paypal accounts.

3. Set up small ($1) and large ($20) monthly subscriptions, and verify that these are put through the correct Paypal accounts.


solarissmoke:
@alanmoo is this a launch blocker? If so would be good if you can test as described above in the next day or so.

alanmoo:
One time payments seem to work, but monthly are failing during the transaction.

This does make me realize there's something I mentioned at one point but I think never made it into a dedicated issue- we should have a minimum dollar amount for transactions, so they don't end up _costing_ us money. (It's $2 USD, I'm not sure about others...I'm looking into the code for the existing site to find it.

solarissmoke:
> One time payments seem to work, but monthly are failing during the transaction.

This should be fixed now - was an issue with all monthly Paypal payments.

> We should have a minimum dollar amount for transactions.

OK - this will require some functional and front-end changes, so the sooner we can specify the details the better.

solarissmoke:
@alanmoo I've tried to debug the issue with monthly payments. We get a failure from Braintree, whose error details are just: "PayPal Validation Failed". There is no error code or more detailed error message, which makes it very difficult for me to work out what is wrong. 

I suspect it is an error raised by Paypal which Braintree doesn't handle - e.g., perhaps it's not possible to set up subscriptions on a micropayments account? This is just a guess. I think you are going to have to engage Braintree to see if they are able to provide more insight.

alanmoo:
I reached out to Braintree today, and am waiting to hear back.

donate/payments/views.py
```diff
@@ -326,8 +325,10 @@ def form_valid(self, form, send_data_to_basket=True):
 
             # Create a subcription against the payment method
             result = gateway.subscription.create({
-                'plan_id': self.get_plan_id(self.currency),
-                'merchant_account_id': self.get_merchant_account_id(self.currency),
+                'plan_id': get_plan_id(self.currency),
+                'merchant_account_id': get_merchant_account_id_for_paypal(
+                    self.currency, form.cleaned_data['amount']
+                ),
```
gideonthomas: Ok, found the problem: we need to pass in the `'first_billing_date'` and that gets it to work for micropayments. Not sure why it already works for macro payments since it goes through the same logic though.

donate/payments/views.py
```diff
@@ -326,8 +325,10 @@ def form_valid(self, form, send_data_to_basket=True):
```
gideonthomas:
```suggestion
start_date = now().date() + relativedelta(months=1)

```

donate/payments/views.py
```diff
@@ -326,8 +325,10 @@ def form_valid(self, form, send_data_to_basket=True):
...
             # Create a subcription against the payment method
             result = gateway.subscription.create({
-                'plan_id': self.get_plan_id(self.currency),
-                'merchant_account_id': self.get_merchant_account_id(self.currency),
+                'plan_id': get_plan_id(self.currency),
+                'merchant_account_id': get_merchant_account_id_for_paypal(
+                    self.currency, form.cleaned_data['amount']
+                ),
                 'payment_method_token': payment_method.token,
                 'price': form.cleaned_data['amount'],
```
gideonthomas:
```suggestion
                'first_billing_date': start_date,
                'price': form.cleaned_data['amount'],
```

donate/payments/views.py
solarissmoke: This would start the subscription one month from today, instead of today? Is that desired behaviour @alanmoo? it's what we do for upsells, but I think that for monthly donations (not upsell) you'd want to start it immediately?

alanmoo: That's correct, we want monthly donations to start immediately.

alanmoo:
This seems to work, but I'm still not seeing Paypal transactions hit Salesforce. (Though that's the case on staging too, so I'm not sure they're related)

solarissmoke:
I've updated the PR to include `first_billing_date`. Paypal issue is fixed in #516.

alanmoo:
While I haven't tested every single configuration (campaigns, regular page, micro, macro) it seems like it's working (I did test micro/macro)

PR #408, opened 2019-10-04
Allow query argument driven override of preset donation amounts.

### LINKED ISSUE #218: Query params for custom donation amounts
It was mentioned in the call the other day and I've discovered that there is still a very valid use case for supporting query parameters to customize the suggested donation amounts: We want the ability to send emails segmented by past donation amounts, with the ability to suggest amounts based on past gifts. This gets much clumsier to do when a separate campaign page needs to be spun up for each one.

solarissmoke:
Closes #218.

@alanmoo grateful if you can test this on the review app and confirm that it works the way you want.

Some examples:

https://donate-wagtail-staging-pr-408.herokuapp.com/en-US/?presets=10,9,8,7
https://donate-wagtail-staging-pr-408.herokuapp.com/en-US/?presets=10,9,8,7&frequency=monthly
https://donate-wagtail-staging-pr-408.herokuapp.com/en-US/?presets=1,9,3,4 - the amounts get sorted descending
https://donate-wagtail-staging-pr-408.herokuapp.com/en-US/?presets=10,9,8,7,34,7,8 - limited to max 4 choices

### DIFF
--- a/donate/core/models.py
+++ b/donate/core/models.py
@@ -1,5 +1,5 @@
 from copy import deepcopy
-from decimal import Decimal
+from decimal import Decimal, InvalidOperation
 
 from django.conf import settings
 from django.db import models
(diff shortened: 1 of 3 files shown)


solarissmoke:
I'll look into the error next week - yes, we can definitely ensure it doesn't 503.

PR #426, opened 2019-10-08
Google Analytics integration.

solarissmoke:
WIP Google Analytics integration, for Chris to test/verify before I finish up tests and tidy up.

Chris, this is using the GA ID `UA-49796218-32`. You may need to configure a view that doesn't filter out our review app domains. You can test at https://donate-wagtail-staging-pr-426.herokuapp.com/card/single/ . 

All the e-commerce tracking should be in place, so it's a case of checking that that works and the correct params are being supplied to GA.

For events, I have left notes on the spreadsheet indicating what is in place, and what needs discussion/clarification.

### DIFF
--- a/app.json
+++ b/app.json
@@ -89,12 +89,12 @@
     "XSS_PROTECTION": "True",
     "CSRF_COOKIE_SECURE": "True",
     "SESSION_COOKIE_SECURE": "True",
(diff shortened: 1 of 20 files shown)


alanmoo:
@philwakefield Based on your slack message earlier today, this is ready?

Issue #438, opened 2019-10-10
Enforce minimum donation amounts for all currencies


solarissmoke:
As per [comment from @alanmoo](https://github.com/mozilla/donate-wagtail/pull/398#issuecomment-540197486) , we need to enforce minimum donation amounts for all currencies, for single and monthly payments.

PR #439, opened 2019-10-10
Enforce minimum amounts on all amount forms.

solarissmoke:
Refs #438. 

There is one issue here which I cannot find a way around. PayPal [doesn't provide any way for us to do validation](https://github.com/paypal/paypal-checkout-components/issues/139) before their flow is initialised. This means that if you enter an "other amount" that is too low and immediately hit the PayPal button, the flow will start and we have no way to stop it. I have added browser-level validation which will *show* an error on the field, but it will not stop the PayPal modal from loading.

The best workaround I have been able to find is to pass invalid data to PayPal, so that it returns an error.

### DIFF
--- a/source/js/components/currency-selector.js
+++ b/source/js/components/currency-selector.js
@@ -43,6 +43,7 @@ class CurrencySelect {
     // Create arrays for monthly and one off based on data
     var oneOffValues = selectedData.presets.single;
     var monthlyValue = selectedData.presets.monthly;
+    var minAmount = selectedData.minAmount;
     var formatter = new Intl.NumberFormat(this.locale, {
       style: "currency",
       currency: selectedData.code.toUpperCase(),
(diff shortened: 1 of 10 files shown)


source/js/components/currency-selector.js
```diff
@@ -171,20 +173,48 @@ class CurrencySelect {
...
+    // selects a preset amount. Otherwise the browser will fail to validate the form.
+    document
+      .querySelectorAll(
+        ".donation-amount__radio:not([data-other-amount-radio])"
+      )
+      .forEach(input => {
+        input.addEventListener("change", e => {
+          if (input.checked) {
+            otherAmountInputs.forEach(
+              input => (input.value = input.checkValidity() ? input.value : "")
```
Pomax: was this supposed to be `===`? Because as a tertiary based on an assignment return this is very scary =)

PR #452, opened 2019-10-11
Ensure the docker backend actually uses the postgres database

cadecairos:
This PR updates the default configuration for docker so that it actually uses Postgres.


cadecairos:
When this lands, people will need to recreate their container stacks in order for Postgres to be properly initialized with the user/pass/db information.

alanmoo:
Using which commands?

patjouk:
`inv docker_nuke_db`

PR #453, opened 2019-10-11
Implement Sentry on the backend and frontend

### LINKED ISSUE #436: Application monitoring / error logging
I'm not sure what Mozilla's approach to logging and reporting on other projects is, but I think it is worth considering setting up some form of application monitoring and error logging on the site before launch (things like Sentry, Scout).

In particular it will be important to log and monitor Braintree payment errors, so that you have visibility on what the causes of payment failures are and can assess whether there are ways to prevent any of them.

cadecairos:
Fixes #436 

It's the end of my day so I didn't take much time to review my own code, let me know if you see anything questionable and we'll work on it next week.


cadecairos:
oh, travis failure. not unexpected.

PR #466, opened 2019-10-15
Sentry Release information

cadecairos:
This PR will pass the value of `HEROKU_RELEASE_VERSION` to Sentry. A new release point will be created the first time it sees a new value, and errors will all be tagged to specific heroku release versions. This should make future debugging easier.


cadecairos:
FYI, I confirmed that sentry has no problem if DSN/Version are not specified by default. it bails out early when not wired up.

alanmoo:
How do I test this?

cadecairos:
If you add the new variables to your env file you can check they're added to the compiled frontend code with your dev tools. I don't think we need to actually test that the packages talk to sentry properly, we can safely assume it works as intended there.

PR #486, opened 2019-10-27
Send campaign to basket

### LINKED ISSUE #480: Campaign not being associated with Contact Record
No test contacts are being associated in Salesforce like they do with petition signatures. CampaignID 701R0000000itrhIAA

(Firefox, Windows 10, Salesforce Staging)

Expected behavior: When using a campaign ID in wagtail, it gets passed to Salesforce.

alanmoo:
Fixes #480, but this won't land on Salesforce until https://github.com/mozmeao/basket/issues/364 is fixed.

### DIFF
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -42,6 +42,7 @@ def send_transaction_to_basket(data):
             'event_type': 'donation',
             'last_name': data['last_name'],
             'first_name': data['first_name'],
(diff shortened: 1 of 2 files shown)


alanmoo:
I've confirmed this passes the campaign ID along with the rest of the transaction data, I just want a sanity check that this won't make some use case fail.

PR #499, opened 2019-10-28
[WIP] Alternate TB copy

TheoChevalier:
Hey @alanmoo, what do you think of this approach?

I’m using a context processor to access the value in templates, but happy to consider a template tag instead if you think it’s more appropriate. Or find a way where we don’t need accessing the value in the template at all.

This PR doesn’t contain all the copy changes needed, I wanted to validate the approach first.


alanmoo:
@solarissmoke this is 100% work we need to take on, not you. But does anything jump to mind insofar as an approach to override the non-Wagtail-tree pages on other instances of the site?

solarissmoke:
I'd propose a different approach that I think will be simpler and doesn't require context switching in code: 

Create a new app in the project (e.g., `thunderbird`), that is conditionally added to `INSTALLED_APPS` based on environment. 

In that app's `templates` directory supply all the templates you wish to override (override specific blocks where needed - you may need to add additional blocks to the base templates to make this as DRY as possible). Ensure that this app is installed at the top of the list so that it takes precedence over existing templates, and then these will be loaded when the app is installed.

(Let me know if this approach needs further elaboration).

alanmoo:
That makes a ton of sense, thank you!

Pomax:
Following up on @solarissmoke's idea, I've filed https://github.com/mozilla/donate-wagtail/pull/589 which introduces that new app with template overrides and conditional loading in settings.py

PR #500, opened 2019-10-28
Implement django-configurations

cadecairos:
~**Marking as a launch blocker because we need this to ensure we don't leak PII in server logs on production.**~

In order to better support differing configurations in dev/stage/prod I'm going to implement [django-configurations](https://django-configurations.readthedocs.io/en/stable/).

This is still WIP since I haven't checked the tests (100% chance they're failing) As of EOD Oct 30th I've just completed coding and a single successful startup of the app.


cadecairos:
@patjouk not quite ready yet, but let me know if you think this approach works for you.

donate/settings/base.py
```diff
@@ -0,0 +1,386 @@
...
+)
+
+
+class Base(Configuration):
+    WAGTAIL_SITE_NAME = "donate"
+    RECAPTCHA_ENABLED = False
+
+    ALLOWED_HOSTS = env('ALLOWED_HOSTS')
+
+    @property
```
cadecairos: these `@property` decorated methods act as computed properties for django settings that need to be evaluated when accessed, or after class declaration.

donate/settings/base.py
```diff
@@ -0,0 +1,386 @@
...
+    def CSP_STYLE_SRC(self):
+        return env('CSP_STYLE_SRC', default=self.CSP_DEFAULT)
+
+    @property
+    def CSP_WORKER_SRC(self):
+        return env('CSP_WORKER_SRC', default=self.CSP_DEFAULT)
+
+    @classmethod
```
cadecairos: django-configurations provides three hooks for configuration classes, `pre_setup`, `setup`, and `post_setup`, where you can add code that needs to execute during the loading of your django settings.

alanmoo:
@cadecairos There's some conflicts

README.md
```diff
@@ -78,8 +78,22 @@ When the Django server is running, you can start the Django shell with:
...
 
+## Customizing your Configuration
+
+You can create a file at `donate/settings/` called `local.py` which will be ignored by git. This file can define a custom
+class for configuring Django/Wagtail. 
+
+```python
+from donate.settings import Base
+
+class MyConfig(Base):
```
alanmoo:
```suggestion

class MyConfig(Base):
```

README.md
```diff
@@ -78,8 +78,22 @@ When the Django server is running, you can start the Django shell with:
...
+    # Add django settings here
+    pass
```
alanmoo:
```suggestion
    pass
    
```

README.md
alanmoo: To make copy/paste into a new file not fail linting.

donate/settings/staging.py
```diff
@@ -0,0 +1,20 @@
...
+class Staging(Base, Secure, Database, Redis, S3, Braintree, Sentry):
+
+    SECRET_KEY = env('DJANGO_SECRET_KEY')
+    BASKET_API_ROOT_URL = env('BASKET_API_ROOT_URL')
+    BASKET_SQS_QUEUE_URL = env('BASKET_SQS_QUEUE_URL')
+
+    @classmethod
+    def pre_setup(cls):
+        cls.LOGGING['loggers']['rq.worker']['handlers'] = ['DEBUG']
+        cls.LOGGING['loggers']['rq.worker']['level'] = 'DEBUG'
```
cadecairos: @alanmoo FYI this is what triggers worker queue logs on staging, but not on production

cadecairos:
> KeyError: "Couldn't setup configuration 'donate.settings.LocalConfig':   "

@alanmoo your `.env` should set `DJANGO_CONFIGURATION` to "LocalConfig" not "donate.settings.LocalConfig", and `DJANGO_SETTINGS_MODULE` to "donate.settings"

env.default
```diff
@@ -1,10 +1,12 @@
 # Wagtail config
+DJANGO_CONFIGURATION='Development'
+DJANGO_SETTINGS_MODULE='donate.settings'
 
 ALLOWED_HOSTS=*
 CONTENT_TYPE_NO_SNIFF=True
 CORS_WHITELIST=*
 CORS_REGEX_WHITELIST=
-DATABASE_URL=
+DATABASE_URL=postgres://postgres@postgres:5432/donate
```
alanmoo:
```suggestion
DATABASE_URL=postgres://donate:mozilla@postgres:5432/donate
```

cadecairos:
> I feel like the additional steps to customize a local deployment should be optional, and this should definitely still build without having to create a file that ends up getting git-ignored.

I agree. I think the default `Development` Configuration needs to be expanded to include the most commonly changed variable and ensure that they can be modified in the .env

I originally added the option for a Local.py to let devs have the option for "total control" over the configuration, but it really shouldn't be necessary to use it for day-to-day work.

alanmoo:
Sounds right. So should we do all of the following?

* Remove any references to local.py/LocalConfig
* Update everything in Development.py to check for the env var, or use the default that's currently on that line (for example, `ALLOWED_HOSTS = '*'` becomes `ALLOWED_HOSTS = env('ALLOWED_HOSTS', default='*')`)
* Remove the defaults from those values in env.default (just make it `ALLOWED_HOSTS=` in .env)

donate/settings/development.py
```diff
@@ -0,0 +1,28 @@
...
+    CORS_WHITELIST = env('CORS_WHITELIST')
+    CORS_REGEX_WHITELIST = env('CORS_REGEX_WHITELIST')
+    SSL_REDIRECT = env('SSL_REDIRECT')
+    USE_X_FORWARDED_FOR = env('USE_X_FORWARDED_FOR')
+    X_FRAME_OPTIONS = env('X_FRAME_OPTIONS')
+    XSS_PROTECTION = env('XSS_PROTECTION')
+    CSRF_COOKIE_SECURE = env('CSRF_COOKIE_SECURE')
+    SESSION_COOKIE_SECURE = env('SESSION_COOKIE_SECURE')
+
+    AWS_LOCATION = 'donate'
```
patjouk: You might want to remove this if you don't use S3 for local dev

donate/settings/environment.py
```diff
@@ -0,0 +1,68 @@
+import os
+import environ
+
+# We set defaults for values that aren't security related
+# to the least permissive setting. For security related values,
+# we rely on it being explicitly set (no default values) so that
+# we error out first.
+env = environ.Env(
+    ALLOWED_HOSTS=(list, '*'),
+    ASSET_DOMAIN=(str, 'donate.mofoprod.net'),
```
patjouk: It's not used on Heroku. Are we're using `AWS_S3_CUSTOM_DOMAIN` instead?

donate/settings/environment.py
```diff
@@ -0,0 +1,68 @@
...
+    AWS_LOCATION=(str, ''),
+    AWS_ACCESS_KEY_ID=(str, ''),
+    AWS_SECRET_ACCESS_KEY=(str, ''),
+    AWS_STORAGE_BUCKET_NAME=(str, ''),
+    AWS_S3_CUSTOM_DOMAIN=(str, ''),
+    AWS_REGION=(str, ''),
+    CONTENT_TYPE_NO_SNIFF=(bool, True),
+    CORS_REGEX_WHITELIST=(tuple, None),
+    CORS_WHITELIST=(tuple, ()),
```
patjouk: I didn't saw those on Heroku either

donate/settings/environment.py
```diff
@@ -0,0 +1,68 @@
...
+    HEROKU_APP_NAME=(str, ''),
+    SET_HSTS=(bool, False),
+    DJANGO_SECRET_KEY=(str, None),
+    SSL_REDIRECT=(bool, False),
+    USE_S3=(bool, False),
+    USE_X_FORWARDED_HOST=(bool, False),
+    USE_X_FORWARDED_FOR=(bool, False),
+    X_FRAME_OPTIONS=(str, 'DENY'),
+    XSS_PROTECTION=(bool, True),
+    REDIS_URL=(str, 'postgres://donate:mozilla@postgres:5432/donate'),
```
patjouk: typo ;)

donate/settings/environment.py
```diff
@@ -0,0 +1,68 @@
...
+    AWS_LOCATION=(str, ''),
+    AWS_ACCESS_KEY_ID=(str, ''),
+    AWS_SECRET_ACCESS_KEY=(str, ''),
+    AWS_STORAGE_BUCKET_NAME=(str, ''),
+    AWS_S3_CUSTOM_DOMAIN=(str, ''),
+    AWS_REGION=(str, ''),
+    CONTENT_TYPE_NO_SNIFF=(bool, True),
+    CORS_REGEX_WHITELIST=(tuple, None),
+    CORS_WHITELIST=(tuple, ()),
+    DATABASE_URL=(str, 'postgres://donate:mozilla@postgres:5432/donate'),
```
patjouk: Do we want the default to be `""` instead?

donate/settings/development.py
```diff
@@ -0,0 +1,28 @@
+from .environment import env
+from .database import Database
+from .base import Base
+from .redis import Redis
+from .braintree import Braintree
+
```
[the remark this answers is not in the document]
alanmoo: I've never used S3 locally; I feel like leaving that in .env is the best approach as it feels like a fair amount of overhead to get it working locally when it's not needed.

donate/settings/base.py
```diff
@@ -0,0 +1,389 @@
...
+    USE_I18N = True
+    USE_L10N = True
+    USE_TZ = True
+    AUTH_USER_MODEL = 'users.User'
+    DJANGO_LOG_LEVEL = env('DJANGO_LOG_LEVEL')
+    ALLOWED_HOSTS = env('ALLOWED_HOSTS')
+    RANDOM_SEED = env('RANDOM_SEED')
+
+    # Initialize these attributes to None
+    BASKET_API_ROOT_URL = None
```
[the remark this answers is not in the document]
cadecairos: yeah there's no reason to define it this way. I'll update

donate/settings/development.py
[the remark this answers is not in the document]
cadecairos: yeah, S3 locally is not a common dev use case, but one we should support, I'll double check to  make sure it's still possible.

donate/settings/development.py
cadecairos: yeah I'll leave this to the S3 configuration mixin

donate/settings/environment.py
cadecairos: Hmm I don't know what this is for. I'll dig a little and find out if we depend on it

donate/settings/environment.py
cadecairos: Any reason in particular we'd want no default?

donate/settings/environment.py
cadecairos: This looks liek a relic from the previous network-api repository merge.

donate/settings/development.py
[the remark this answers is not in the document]
cadecairos: should be better now

cadecairos:
I've updated a bunch of things based on the feedback I've received so far. There have been other changes made when I found certain issues, such as how we've been using environ incorrectly this whole time, and the Thunderbird configuration classes in lieu of using a boolean flag.

donate/settings/environment.py
```diff
@@ -0,0 +1,95 @@
+import os
+import environ
+
+CSP_DEFAULT_SRC = ("'self'")
+CSP_SCRIPT_SRC = ("'self'", '\'unsafe-inline\'', '\'unsafe-eval\'',
```
Pomax: if `"'...'"` works, we should probably use that for the rest of the settings too, rather than `'\'...\''`.

donate/settings/environment.py
```diff
@@ -0,0 +1,95 @@
...
+    USE_X_FORWARDED_FOR=(bool, False),
+    USE_X_FORWARDED_HOST=(bool, False),
+    WAGTAILLOCALIZE_PONTOON_GIT_CLONE_DIR=(str, ''),
+    WAGTAILLOCALIZE_PONTOON_GIT_URL=(str, ''),
+    X_FRAME_OPTIONS=(str, 'DENY'),
+    XSS_PROTECTION=(bool, True),
+)
+
+
+app = environ.Path(__file__) - 2
```
Pomax: is this... arithmetics on paths? O_O

donate/settings/production.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @classmethod
+    def pre_setup(cls):
+        # production forces debug to be False, so the rq worker will never log donor data with the config below.
+        cls.LOGGING['loggers']['rq.worker']['handlers'] = ['DEBUG']
+        cls.LOGGING['loggers']['rq.worker']['level'] = 'DEBUG'
+
+
+class ThunderbirdProduction(Production, ThunderbirdOverrides):
+    INSTALLED_APPS = ThunderbirdOverrides.INSTALLED_APPS + Production.INSTALLED_APPS
```
Pomax: this is cool

donate/settings/production.py
```diff
@@ -0,0 +1,32 @@
...
+
+    @property
+    def TEMPLATES(self):
+        config = Production.TEMPLATES
+        config[0]['DIRS'] = ThunderbirdOverrides.TEMPLATES_DIR + config[0]['DIRS']
```
Pomax: this is also cool

donate/settings/staging.py
```diff
@@ -0,0 +1,31 @@
...
+    BASKET_API_ROOT_URL = env('BASKET_API_ROOT_URL')
+    BASKET_SQS_QUEUE_URL = env('BASKET_SQS_QUEUE_URL')
+
+    @classmethod
+    def pre_setup(cls):
+        cls.LOGGING['loggers']['rq.worker']['handlers'] = ['info']
+        cls.LOGGING['loggers']['rq.worker']['level'] = 'INFO'
+
+
+class ThunderbirdStaging(Staging, ThunderbirdOverrides):
```
Pomax: Is there a way to make the ThunderbirdOverrides automatically hook into the `INSTALLED_APPS` and `TEMPLATES` variables? Or does it not have access to "the object that's been created so far" in a multiple inheritance chain?

donate/settings/environment.py
cadecairos: hahaha that was my reaction. but apparently it works. it's _magic_

donate/settings/staging.py
cadecairos: I initially tried that, but django-configurations doesn't seem to work that way. The solution I ended up with is the first combination of configurations that actually did what I was trying to do.

donate/settings/environment.py
alanmoo: skeptical face

cadecairos: I was curious, this is a `Path` class defined in django-environ: https://github.com/joke2k/django-environ/blob/develop/environ/environ.py#L735-L749

cadecairos:
@patjouk @alanmoo I believe I've addressed your reviews, could you both take another look so we can merge this and validate it in our various environments?

donate/settings/base.py
```diff
@@ -0,0 +1,373 @@
...
+
+    # Domain Redirects
+    DOMAIN_REDIRECT_MIDDLEWARE_ENABLED = env('DOMAIN_REDIRECT_MIDDLEWARE_ENABLED')
+    TARGET_DOMAINS = env('TARGET_DOMAINS')
+
+    # Initialize these attributes to None
+    BASKET_API_ROOT_URL = env('BASKET_API_ROOT_URL')
+    BASKET_SQS_QUEUE_URL = env('BASKET_SQS_QUEUE_URL')
+
+    # Pontoon settings
```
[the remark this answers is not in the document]
cadecairos: RE: Thunderbird, I've ported that functionality over to specific configuration classes for Thunderbird, rather than a bool flag.

donate/settings/thunderbird.py
```diff
@@ -0,0 +1,6 @@
+from .environment import app
+
+
+class ThunderbirdOverrides(object):
+    INSTALLED_APPS = ['donate.thunderbird']
+    TEMPLATES_DIR = [app('thunderbird/templates')]
```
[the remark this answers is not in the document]
cadecairos: We don't need to include a boolean, since using the Thunderbird configurations implicitly implies that `THUNDERBIRD_INSTANCE` is set to True

cadecairos:
remaining todo:

- [ ] Fix default Braintree config
- [ ] Review app webhook env vars in base.py and environment.py
- [ ] Verify review apps are working under this branch

cadecairos:
>  Fix default Braintree config

I'm pretty sure what this PR does doesn't change the existing behaviour out of the box. I'd like to handle that in another issue if it's a problem.

.travis.yml
```diff
@@ -42,3 +41,4 @@ matrix:
     - nvm use --lts=carbon
     install: "./travis-scripts/npm-install.sh"
     script: npm run test
+
```
Pomax: 😄

donate/settings/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from .redis import Redis
+from .secure import Secure
+from .sentry import Sentry
+from .testing import Testing
+from .development import Development, ThunderbirdDevelopment
+from .review_app import ReviewApp
+from .staging import Staging, ThunderbirdStaging
+from .production import Production, ThunderbirdProduction
+
+__all__ = [
```
Pomax: Django has so much magic in it, I really don't understand why there is no way to just tell it to imexport everything in the dir...

donate/settings/base.py
```diff
@@ -0,0 +1,385 @@
...
+    # Static content settings
+    STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'
+    STATICFILES_DIRS = [app('frontend')]
+    STATIC_ROOT = root('static')
+    STATIC_URL = '/static/'
+
+    MEDIA_ROOT = root('media/')
+    MEDIA_URL = '/media/'
+
+    HEROKU_APP_NAME = ''
```
Pomax: This feels like it wants to be an env var? (with `'''` fallback?)

donate/settings/base.py
```diff
@@ -0,0 +1,385 @@
...
+
+    # Review apps' slack bot
+    GITHUB_TOKEN = env('GITHUB_TOKEN')
+    SLACK_WEBHOOK_RA = env('SLACK_WEBHOOK_RA')
+
+    # Pontoon check slack bot
+    SLACK_WEBHOOK_PONTOON = env('SLACK_WEBHOOK_PONTOON')
+    TRAVIS_LOGS_URL = env('TRAVIS_JOB_WEB_URL', default='')
+
+    @property
```
Pomax: this is great

donate/settings/base.py
```diff
@@ -0,0 +1,385 @@
...
+        },
+
+        # Must be a separate queue as it's limited to one item at a time
+        'wagtail_localize_pontoon.sync': {
+            'URL': 'redis://redis:6379/0',
+            'DEFAULT_TIMEOUT': 500,
+        },
+    }
+
+    LANGUAGES = [
```
Pomax: a nit, but would it be possible to make this a dedicated list in something like `languages.py` and then importing from there?

donate/settings/development.py
```diff
@@ -0,0 +1,61 @@
...
+    XSS_PROTECTION = env('XSS_PROTECTION')
+    CSRF_COOKIE_SECURE = env('CSRF_COOKIE_SECURE')
+    SESSION_COOKIE_SECURE = env('SESSION_COOKIE_SECURE')
+
+    # In dev, lets not worry about enforcing password changes
+    AUTH_PASSWORD_VALIDATORS = env('AUTH_PASSWORD_VALIDATORS')
+
+    @classmethod
+    def pre_setup(cls):
+        super(Development, cls).pre_setup()
```
Pomax: We're in python 3, so the simpler `super().pre_setup()` should work (and I guess the same for any other super(...) calls)

donate/settings/environment.py
```diff
@@ -0,0 +1,113 @@
+import os
+import environ
+
+CSP_DEFAULT_SRC = ("'self'")
+CSP_SCRIPT_SRC = ("'self'", "'unsafe-inline'", "'unsafe-eval'",
```
Pomax: Point of concern: do we _need_ `unsafe-inline` and `unsafe-eval`? They're kind of... unsafe

donate/settings/secure.py
```diff
@@ -0,0 +1,38 @@
...
+
+class Secure(object):
+    DEBUG = False
+    USE_X_FORWARDED_HOST = True
+    CSRF_COOKIE_SECURE = True
+    SESSION_COOKIE_SECURE = True
+    SECURE_BROWSER_XSS_FILTER = True
+    SECURE_CONTENT_TYPE_NOSNIFF = True
+    SECURE_HSTS_INCLUDE_SUBDOMAINS = True
+    SECURE_HSTS_SECONDS = 60 * 60 * 24 * 31 * 6
```
Pomax: I think the recommended value is 1 year? (also, fun aside: six months is not 186 days =D)

donate/settings/base.py
cadecairos: This is really only ever needed when using the `ReviewApp` configuration, so I can probably drop it from base.py altogether.

donate/settings/base.py
cadecairos: yeah I'm okay with that.

donate/settings/environment.py
cadecairos: I checked as part of the VA and we determined that we do need at least one of them I'll check which it is.

donate/settings/base.py
cadecairos: yes

donate/settings/environment.py
cadecairos: Based on my tests, we must use unsafe-inline because of Paypal. We can remove unsafe-eval.

donate/settings/environment.py
```diff
@@ -0,0 +1,113 @@
...
+CSP_CONNECT_SRC = ("'self'", 'api.sandbox.braintreegateway.com',
+                   'client-analytics.sandbox.braintreegateway.com', 'api.braintreegateway.com',
+                   'client-analytics.braintreegateway.com', '*.braintree-api.com', 'www.paypal.com',
+                   'www.google-analytics.com', 'https://www.mozilla.org/en-US/newsletter/',
+                   'https://sentry.prod.mozaws.net')
+CSP_BASE_URI = CSP_DEFAULT_SRC
+CSP_WORKER_SRC = CSP_DEFAULT_SRC
+
+BRAINTREE_MERCHANT_ACCOUNTS = {
+    'usd': 'usd-ac',
```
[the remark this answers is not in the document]
cadecairos: I don't even know where this came from. might be what Torchbox had by default

donate/settings/braintree.py
```diff
@@ -0,0 +1,19 @@
...
+
+class Braintree(object):
+    BRAINTREE_USE_SANDBOX = env('BRAINTREE_USE_SANDBOX')
+    BRAINTREE_MERCHANT_ID = env('BRAINTREE_MERCHANT_ID')
+    BRAINTREE_PUBLIC_KEY = env('BRAINTREE_PUBLIC_KEY')
+    BRAINTREE_PRIVATE_KEY = env('BRAINTREE_PRIVATE_KEY')
+    BRAINTREE_TOKENIZATION_KEY = env('BRAINTREE_TOKENIZATION_KEY')
+    BRAINTREE_MERCHANT_ACCOUNTS = env('BRAINTREE_MERCHANT_ACCOUNTS')
+    BRAINTREE_PLANS = env('BRAINTREE_PLANS')
+    BRAINTREE_MERCHANT_ACCOUNTS_PAYPAL_MICRO = env('BRAINTREE_MERCHANT_ACCOUNTS_PAYPAL_MICRO')
```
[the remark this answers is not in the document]
cadecairos: `environment.py` defaults this to `{}` so I think so?

PR #505, opened 2019-10-30
Refactor sending of subscription transactions to Basket

### LINKED ISSUE #216: Pass data from Braintree webhooks to Basket
As per discussions on slack there is a need to collect some information from Braintree webhooks and relay this to Basket. 

## Information required

1. What information needs to be collected? For subscriptions, Braintree can send [a variety of different kinds of notifications](https://developers.braintreepayments.com/reference/general/webhooks/subscription/python#notification-kinds). Which of these are we interested in?

- Subscription cancelled.
- Subscription charged successfully.
- Subscription charged unsuccessfully.
- Subscription expired.

2. What is the data structure expected by Basket for each of the different kinds of event that we need to listen to?

### LINKED ISSUE #459: Upsell transaction record is being created even though transaction status is "Pending"
For Paypal and CC upsells, these records should not land in SFDC until the transaction status is Active or SFDC will record the transaction as stage = Closed Won and Close Date as date the subscription was created but not charged.

(BT View)
2019_10_14_BT_Subscription_Pending

SFDC View
2019_10_14_SFDC_Subscription_Pending

(Windows 10, Firefox and Chrome)

solarissmoke:
Fixes #216, fixes #459.

- Don't send subscription initialisation information to Basket.
- Add webhook processing for successful subscription charges.
- Pass `locale` to Braintree as a custom field.

The webhook will need a bit of testing as the Braintree provides limited testing data. In particular I don't know that custom fields that are stored against a customer when creating a subscription are then passed to the custom fields for each transaction (and if they aren't, I can't currently see how we get that data).

### DIFF
--- a/donate/payments/views.py
+++ b/donate/payments/views.py
@@ -32,6 +32,7 @@ class BraintreePaymentMixin:
         return {
             'project': self.request.session.get('project', 'mozillafoundation'),
             'campaign_id': self.request.session.get('campaign_id', ''),
+            'locale': self.request.LANGUAGE_CODE,
         }
 
     def get_merchant_account_id(self, currency):
--- a/donate/payments/views.py
+++ b/donate/payments/views.py
@@ -176,7 +176,7 @@ class CardPaymentView(BraintreePaymentMixin, FormView):
         if self.payment_frequency == constants.FREQUENCY_SINGLE:
             return self.process_single_transaction(form, send_data_to_basket=send_data_to_basket)
         else:
-            return self.process_monthly_transaction(form, send_data_to_basket=send_data_to_basket)
+            return self.process_monthly_transaction(form)
 
     def create_customer(self, form):
         result = gateway.customer.create({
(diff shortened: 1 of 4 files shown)


solarissmoke:
The webhook processing is hard to test until we merge so that webhooks are received from Braintree. I guess worth testing that the upsell txs are no longer being sent before we do that?

donate/payments/views.py
```diff
@@ -257,7 +258,7 @@ def process_monthly_transaction(self, form, send_data_to_basket=True):
                 form,
                 transaction_id=result.subscription.id,
                 last_4=payment_method.last_4,
-                send_data_to_basket=send_data_to_basket,
+                send_data_to_basket=False,
```
alanmoo: @solarissmoke  I'm not sure if we should be doing this or not. When creating a monthly transaction that processes immediately, does the webhook still get fired? Or is this one of those things we just kind of need to find out with Braintree?

@stephaniemcv if someone makes a recurring donation (not an upsell), the first transaction should post immediately to basket, correct?

solarissmoke: The webhook still gets fired. At this stage all we have is a _subscription_, and no transactions have been created against it. The first transaction is not created immediately the subscription is created.

Issue #488, opened 2019-10-27
Post-donation newsletter signup doesn't actually trigger a specific newsletter signup


alanmoo:
After making a donation, the view that asks if a user wants to sign up for the newsletter submits properly to Basket, but the donor doesn't actually get signed up for a specific list.

alanmoo:
This was fixed by #510

PR #510, opened 2019-10-30
Fixing newsletter POST URL

alanmoo:
I think this takes care of #488 but given the way Redis works right now I need to merge to find out. But since it's broken anyway, 🤷‍♂ yolo?

### DIFF
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -31,7 +31,7 @@ def send_newsletter_subscription_to_basket(data):
         'email': data['email']
     }
     try:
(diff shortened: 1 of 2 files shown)


donate/payments/tasks.py
```diff
@@ -26,12 +26,11 @@ def send_newsletter_subscription_to_basket(data):
         'format': 'html',
         'lang': data['lang'],
         'newsletters': 'mozilla-foundation',
-        'trigger_welcome': 'N',
```
alanmoo: This is apparently configured on the SFDC side now.

PR #512, opened 2019-10-31
[Pontoon] Get the SSH key during the release step

patjouk:
We're using Heroku pipeline, which means that when an app is `promoted` to prod, [it's not rebuild](https://devcenter.heroku.com/articles/pipelines#design-considerations). The SSH buildback is never executed on prod: if you check the SSH key from `id_rsa` on prod, you will see that it's the one from staging.

I'm moving the `id_rsa` file creation to `release steps.sh`, which is always executed during the deployment process.


patjouk:
Won't work.

PR #516, opened 2019-11-01
Fix generation of session data for Paypal transactions.

### LINKED ISSUE #515: Paypal transactions aren't getting to Salesforce
One-time donations are the only ones working right now (see #514) and they show up in Braintree, but not Salesforce.

solarissmoke:
Fixes #515.

@alanmoo the tests didn't catch this because I had split the tests for (a) generating the session data and (b) sending this to basket - and had omitted to include integration tests to link up the two. I have addressed this now so that the tests check that the data passes correctly all the way through.


alanmoo:
Great, this sent data to staging!

solarissmoke:
You should see names in salesforce as we are fetching them from the API. I don't know why Braintree dashboard doesn't show names for PayPal transactions. There may be a privacy issue but then the API does give us this information.

PR #538, opened 2019-11-05
Fix client-side handling of Paypal monthly payments when recaptcha is enabled

### LINKED ISSUE #514: Monthly Paypal transactions are failing
This may be fixed by #398 but Heroku's having issues today.

Attempting to make a monthly Paypal donation on staging (no matter the size) fails.

solarissmoke:
Fixes #514.


alanmoo:
Ah, this makes sense as to why it was working on single transactions. Thank you!

PR #539, opened 2019-11-05
Add subscription ID to webhooks

### LINKED ISSUE #456: Recurring subscription id recording as transaction id
For both Paypal and CC recurring charges, Salesforce is recording no value in the PMT Subscription ID field and is instead recording the BT Subscription ID value in the PMT Transaction ID field

BrainTree view
2019_10_14_BT_SubID_recording_as_TranID

SFDC view
2019_10_14_SFDC_SubID_recording_as_TranID

(Windows 10/Chrome)

alanmoo:
Closes #456

### DIFF
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -85,6 +85,7 @@ class BraintreeWebhookProcessor:
             'payment_frequency': constants.FREQUENCY_MONTHLY,
             'payment_method': constants.METHOD_PAYPAL if is_paypal else constants.METHOD_CARD,
             'transaction_id': last_tx.id,
+            'subscription_id': notification.subscription.id,
             'project': custom_fields.get('project', ''),
             'last_4': None if is_paypal else last_tx.credit_card_details.last_4,
             'landing_url': '',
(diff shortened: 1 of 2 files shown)


alanmoo:
@solarissmoke I'm mainly looking for a "this makes sense" from you.

donate/payments/tasks.py
```diff
@@ -85,6 +85,7 @@ def process_subscription_charged_successfully(self, notification):
             'payment_frequency': constants.FREQUENCY_MONTHLY,
             'payment_method': constants.METHOD_PAYPAL if is_paypal else constants.METHOD_CARD,
             'transaction_id': last_tx.id,
+            'subscription_id': notification.subscription.id,
```
solarissmoke: This won't have any effect on its own - `send_transaction_to_basket` looks for specific keys in the dictionary that is passed to it, and `subscription_id` isn't one of them - we need to modify that function as well to pass through this parameter.

alanmoo: Ah, thanks!

PR #541, opened 2019-11-05
Handle differences in Paypal webhooks

### LINKED ISSUE #540: Paypal recurring subscriptions aren't passing email to Basket
Looks like the details of a paypal payment come in a different format from the Braintree webhook.
I'm opening a PR momentarily.

(FYI @solarissmoke for future work)

alanmoo:
Closes #540 

@gideonthomas Tests are written, I mainly want a Python check here. There's probably ways I could have optimized it but it feels...fine?

### DIFF
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -90,7 +90,7 @@ class BraintreeWebhookProcessor:
                 'email': last_tx.customer_details.email,
             }
         else:
-            logger.error("Unexpected payment type on subscription webhook")
+            logger.error(f"Unexpected payment type on subscription webhook: {last_tx.payment_instrument_type}")
 
         send_transaction_to_basket({
             'last_name': donor_details.get('last_name', ''),
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -90,7 +90,7 @@ class BraintreeWebhookProcessor:
                 'email': last_tx.customer_details.email,
             }
         else:
-            logger.exception("Unexpected payment type on subscription webhook")
+            logger.critical("Unexpected payment type on subscription webhook")
 
         send_transaction_to_basket({
             'last_name': donor_details.get('last_name', ''),
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -90,7 +90,7 @@ class BraintreeWebhookProcessor:
                 'email': last_tx.customer_details.email,
             }
         else:
-            logger.critical("Unexpected payment type on subscription webhook")
+            logger.error("Unexpected payment type on subscription webhook")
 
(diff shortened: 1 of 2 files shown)


donate/payments/tasks.py
```diff
@@ -75,11 +75,28 @@ def process_subscription_charged_successfully(self, notification):
...
+                "email": last_tx.paypal_details.payer_email
+            }
+        elif last_tx.payment_instrument_type == 'credit_card':
+            donor_details = {
+                'last_name': last_tx.customer_details.last_name,
+                'first_name': last_tx.customer_details.first_name,
+                'email': last_tx.customer_details.email,
+            }
+        else:
+            logger.error("Unexpected payment type on subscription webhook")
```
cadecairos:
```suggestion
            logger.error(f"Unexpected payment type on subscription webhook: {last_tx.payment_instrument_type}")
```

donate/payments/tasks.py
```diff
@@ -75,11 +75,28 @@ def process_subscription_charged_successfully(self, notification):
...
+
         send_transaction_to_basket({
-            'last_name': last_tx.customer_details.last_name,
-            'first_name': last_tx.customer_details.first_name,
+            'last_name': donor_details.get('last_name', ''),
+            'first_name': donor_details.get('first_name', ''),
             'campaign_id': custom_fields.get('campaign_id', ''),
-            'email': last_tx.customer_details.email,
+            'email': donor_details.get('email', ''),
```
cadecairos: Curious if the use of `.get()` with a default of `''` is necessary? are we expecting these keys to sometimes be empty?

alanmoo: If email is empty, basket will barf. But I suppose name could be empty at some point. (And we can't use JS-like object notation because it's a dict, not a proper object. I found that out when tests were failing.)

alanmoo: But I don't think we are expecting email to ever be empty in this case.

cadecairos: Well the alternative to `foo.get('bar')` is `foo['bar']` for dicts - which is why I bring it up, since we can just do that if we don't expect to have to set default values.

alanmoo: I could have sworn I'd tried that and it wasn't happy.

gideonthomas: the problem with `foo['bar']` is if `bar` doesn't exist in `foo`, it will throw an exception, whereas `foo.get('bar')` will return `None` without throwing.

cadecairos: It's fine with me either way. If we don't trust that Braintree is going to give us consistent data in webhooks we can use `.get()`

Issue #446, opened 2019-10-10
Cleanup tasks


solarissmoke:
A laundry list of things I think we should try and clean up (if possible, before handover to Mozilla):

- [ ] `wagtail-localize` and `wagtail-localize-pontoon` are currently being installed from github repos that are owned by Karl. The git installation is problematic, because Heroku's build cache doesn't handle them properly, and has to be manually cleared in order to update (even if the tag is changed in Pipfile). These should be set up on pypi.

- [x] Configuration management is a bit clunky at the moment - `env.default` doesn't provide a works-out-of-the-box development configuration. It would be nice if it did, or we provided instructions on where to obtain the missing parameters (e.g, sandbox Braintree credentials).

- [x] CSP, while an important security measure, comes with the risk of breaking stuff e.g., if third parties change their dependencies. I think it is worth considering setting up CSP reporting before enforcing it. Particular concerned about Paypal who seem to do all sorts of funny stuff in their scripts.

- [x] Advanced Fraud Detection is currently mandatory in our code. On reflection I wonder whether we need to make this configurable, so that it can be easily turned off in the event that (for whatever reason) legitimate transactions are being declined). Don't know what the likelihood of this is, and we won't until we are in production, so question is whether to be prepared in case.

alanmoo:
Localize repos: Yes. Tell me what you need me to spin up, if @TheoChevalier and @patjouk haven't done it already.

Configuration management: At least having documentation mentioning to `cp env.default` to some other file is needed. I was able to pretty easily copy values from staging and get things working, so I'm not super concerned about this though if you want to make this more accessible that'd be nice.

CSP: We've got Sentry configured, so I'm less concerned about this as we should catch stuff pretty quickly.

Fraud Detection: Good point. Again, probably more useful for broader uses of this project, but I like the idea of a safety valve.

TheoChevalier:
> Localize repos: Yes. Tell me what you need me to spin up, if @TheoChevalier and @patjouk haven't done it already.

AFAICT, we haven’t, we’ve only created repositories for localization content, not for the wagtail modules source code.

cadecairos:
no real work tasks here for eng team. refile issues individually if work remains

PR #542, opened 2019-11-05
Settings for local dev

patjouk:
Related PR #446
I've updated the settings and invoke commands available for local dev. It's fixing a bunch of issues (`inv docker-nuke-db` replaced by a more resilient task), and it's using the same tasks as the one available on the foundation site.

It will need to be updated when #500 lands.


tasks.py
```diff
@@ -17,100 +17,42 @@
...
-    manage(ctx, "makemessages --keep-pot --no-wrap")
-    manage(ctx, "makemessages -d djangojs --keep-pot --no-wrap --ignore=node_modules")
-    os.replace("donate/locale/django.pot", "donate/locale/templates/LC_MESSAGES/django.pot")
-    os.replace("donate/locale/djangojs.pot", "donate/locale/templates/LC_MESSAGES/djangojs.pot")
+def create_docker_env_file(env_file):
+    """Create or update an .env to work with a docker environment"""
+    with open(env_file, 'r') as f:
+        env_vars = f.read()
+    # update the DATABASE_URL env
+    new_db_url = "DATABASE_URL=postgres://postgres@postgres:5432/postgres"
```
cadecairos: This might change if we land #452

PR #543, opened 2019-11-05
Set the rq.worker logger to DEBUG level

cadecairos:
Setting this logger to DEBUG will ensure it doesn't log on production where the `DJANGO_LOG_LEVEL` is set to `INFO`


cadecairos:
Testing locally isn't possible while we can't run with `DEBUG` set to True, but I tested with the review app and confirmed that PII isn't being logged by the worker.

PR #545, opened 2019-11-05
Subclass whitenoise.storage.CompressedManifestStaticFilesStorage

cadecairos:
... so that manifest_strict can be set to False, allowing debug=False to work with S3 storage disabled.

For this to work, you must run collectstatic. `inv docker-manage "collectstatic --no-input"`

### DIFF
--- a/donate/settings.py
+++ b/donate/settings.py
@@ -356,8 +356,8 @@ django.conf.locale.LANG_INFO['uk'] = {
     'name_local': 'Ukrainian',
 }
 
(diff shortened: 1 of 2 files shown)


cadecairos:
> 1.) This still results in a broken image on the homepage

(when `DEBUG=False`)

cadecairos:
Closing, in favour of a new issue, to investigate who to properly handle static files when `DEBUG=False` in local dev.

PR #548, opened 2019-11-06
Fix locale code in share buttons

TheoChevalier:
Just realized locales with a country code were using the `ab_CD` format in share buttons instead of `ab-CD`, leading to 404s. Didn’t catch that, as I only tested with locales without a country code


alanmoo:
@TheoChevalier I made some changes to the share page the other day, removing locale in the page we share (to allow the recipient's browser to pick it up again). As a result, this PR needs conflicts resolved. I suspect we can remove the calls to these new tags, but I'm not positive on the best l10n approach here.

TheoChevalier:
@alanmoo When fixing https://github.com/mozilla/donate-wagtail/issues/597, if we’re updating links to an external website, say, thunderbird.net, we won’t need this PR at all

alanmoo:
Ah, that makes sense. So I'll just close this one?

PR #552, opened 2019-11-07
Add TargetDomainRedirectMiddleware

cadecairos:
Adds the `DOMAIN_REDIRECT_MIDDLEWARE_ENABLED` and `TARGET_DOMAINS` settings. 

When enabled, Django will redirect requests with a HOST header that don't match the `TARGET_DOMAINS` list to the "primary" domain, i.e. `TARGET_DOMAINS[0]`


cadecairos:
Confirmed working on staging: https://donate-wagtail-staging.herokuapp.com/

Production configuration is in place.

PR #555, opened 2019-11-07
Improve Sentry error collection for Payments

cadecairos:
This PR will continue capturing errors processing Paypal donations which we can update down the line once we're certain it's working.

Based on the [documentation I found](https://developers.braintreepayments.com/reference/response/transaction/python#unsuccessful-result) whenever `result.is_successful` is false, it's either due to form validation issues, or card processing issues, so we shouldn't need to bother capturing error logging here.

### DIFF
--- a/donate/payments/views.py
+++ b/donate/payments/views.py
@@ -1,4 +1,5 @@
 import logging
+from sentry_sdk.integrations.django import ignore_logger
 
 from django.conf import settings
 from django.contrib import messages
@@ -26,7 +27,9 @@ from .utils import (
     get_suggested_monthly_upgrade, freeze_transaction_details_for_session
 )
 
+ignore_logger(__name__)
 logger = logging.getLogger(__name__)
+sentry_logger = logging.getLogger(f'{__name__}_sentry')
 
 
 class BraintreePaymentMixin:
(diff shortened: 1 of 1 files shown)


donate/payments/views.py
```diff
@@ -393,9 +396,10 @@ def form_valid(self, form, send_data_to_basket=True):
                 payment_method = result.customer.payment_methods[0]
                 success_kwargs['payment_method'] = payment_method
             else:
-                logger.error(
+                sentry_logger.error(
                     'Failed to create Braintree customer: {}'.format(result.message),
-                    extra={'result': result}
+                    extra={'result': result},
+                    exc_info=True
```
cadecairos: includes stack trace.

donate/payments/views.py
```diff
@@ -393,9 +396,10 @@ def form_valid(self, form, send_data_to_basket=True):
                 payment_method = result.customer.payment_methods[0]
                 success_kwargs['payment_method'] = payment_method
             else:
-                logger.error(
+                sentry_logger.error(
```
cadecairos: This is inside the PaypalPaymentView

cadecairos:
@alanmoo / @patjouk I think this is ready for a review.

PR #576, opened 2019-11-12
Disable amex for all but USD

### LINKED ISSUE #560: Update supported currencies to remove Amex logo for most
Since we’re only accepting Amex for USD, (right @gideonthomas?) we should update [this file](https://github.com/mozilla/donate-wagtail/blob/master/donate/payments/constants.py) to indicate that.

alanmoo:
Closes #560


alanmoo:
This is just syntax changes ultimately, so since it passes tests and works right I'm gonna land it.

PR #577, opened 2019-11-13
Refactor accordion block to use a StreamBlock.

### LINKED ISSUE #518: Accordion content is not extracted in .po files
I’ve added an accordion component with a few question/answers for testing purposes on https://donate-wagtail.mofostaging.net/en-US/faq/ then synced content with the Pontoon repo, and only the page title + CTA got synced: https://github.com/kaedroho/mozilla-donate-staging-content/commit/0d1db03f372f6859cd45170a05338dd4f2e8115b

solarissmoke:
Fixes #518.

@kaedroho I have refactored the model to use a StreamBlock - unfortunately this doesn't yet work with wagtail-localize because [`StreamFieldSegmentExtractor`](https://github.com/kaedroho/wagtail-localize/blob/6b4426dbac8fb4e1201e71e8cabe64830ecf8019/wagtail_localize/segments/extract.py#L15) doesn't currently handle StreamBlocks that are not directly the child of a StreamField. 

I think it is easier to fix this in wagtail-localise rather than supplying a `get_translatable_segments()` method here. In either case I think it will be much quicker for you to resolve!


kaedroho:
Fixed on master now

kaedroho:
Just released `wagtail-localize` 0.3.2 which fixes that issue

solarissmoke:
Hey @alanmoo this refactors the accordion blocks so that they can be translated. One caveat is that the migration will empty all existing accordions because the data structure has changed.

TheoChevalier:
Thanks both! And don’t worry about existing data, it’s not too long to repopulate

PR #592, opened 2019-11-20
case sensitivity

Pomax:
Does not break on macos, due to filesystem case insensitivity. Definitely breaks on a well-configured linux/unix machine.


alanmoo:
Sentry issue: [DONATE-STAGE-W](https://sentry.prod.mozaws.net/operations/donate-stage/issues/6697724/)

PR #621, opened 2019-11-27
add credit card type to our data reporting

### LINKED ISSUE #620: Capture credit card type in Salesforce
According to Abigail, we need to include the credit card type on the receipt emailed to donors. This means we need to send the card type to Basket, which, according to @stephaniemcv, is already set up to handle this.

We'll need to make sure we're passing it in for both one-time and monthly credit card transactions.

Pomax:
Closes #620

This adds a `card_type` to our `CardPaymentView.get_transaction_details_for_session` view function, which will take on the value "None" (that's a string, not the python `None`) if for some reason there is no credit card information in the transaction, or any of the values listed in https://developers.braintreepayments.com/reference/response/transaction/python#credit_card_details.card_type


Pomax:
@lucie would you happen to know how I can update the tests? It looks like the function I update gets tested against a monthly-subscription upsell, but if I test this "For real" by making a donation with the test Visa credit card number and then clicking "yes" on the request to become a monthly donor, the function doesn't seem to get called, so I might be missing something important here =S

Pomax:
Note that I did add some mocking to the structure that I should be seeing in the function that's doing the extraction (a result object with a `.transaction` property), but the tests are failing with the report about a completely different structure is being passed in (a subscription).

Pomax:
@patjouk @alanmoo tests now pass, I'm not 100% confident I did this right, so review feels more important than usual =D

donate/payments/views.py
```diff
@@ -329,12 +329,27 @@ def process_monthly_transaction(self, form):
             return self.process_braintree_error_result(result, form)
 
     def get_transaction_details_for_session(self, result, form, **kwargs):
+        card_type = kwargs.get('card_type', 'Unknown')
+
+        if card_type == 'Unknown':
```
solarissmoke: I would strongly urge against having in here that runs only during testing - this is liable to introduce bugs in production. 

I think the issue is that you're not actually supplying a `card_type` kwarg anywhere, so this kwarg will never be set (see comment above). The mock result classes used in the tests may need to be modified so that they supply this new property in the appropriate places.

donate/payments/views.py
```diff
@@ -329,12 +329,27 @@ def process_monthly_transaction(self, form):
             return self.process_braintree_error_result(result, form)
 
     def get_transaction_details_for_session(self, result, form, **kwargs):
+        card_type = kwargs.get('card_type', 'Unknown')
```
solarissmoke: It doesn't look like `card_type` is every being passed to the kwargs? This needs to be supplied in the calls to `self.success()` that happen in `process_single_transaction` and `process_monthly_transaction`.

solarissmoke:
@alanmoo short on time today, but have left a couple of comments. Will have a bit of time tomorrow if you want me to look closer/run locally.

donate/payments/views.py
Pomax: This function runs fine during normal operation, where this value should be `Unknown` unless the transaction we get from Braintree includes an explicit card_type (it might not, so I want to make doubly sure we have that fallback value set up first).

During normal operation, there is no `card_type` in `kwargs`, ever, as far as I can tell. That only happens during testing, but when it happens during testing, not pulling the value from `kwargs` causes test errors, and this code is complex enough that I literally have no idea _why_ the tests call this function in they way they do.

Can you expand on what other changes would be necessary to the mocked structures? Because I've already added it to the only structure that has `card_type` as far as can be found in Braintree's API documentation (https://github.com/mozilla/donate-wagtail/blob/c60888644da6034efa7252463c97dde9c87a88c7/donate/payments/tests/test_views.py#L39)

donate/payments/views.py
Pomax: It's not, because this is the function where card_type, as one of the transaction details, gets extracted from Braintree's `result` transaction.  The only reason this line is `kwargs.get('card_type', 'Unknown')` instead of `= 'Unknown'` is because for some reason, two tests for "monthly" put their own `card_type` in `kwargs` (something I don't see happening during real monthly upsell user actions) and so if this function does not check whether something has put `card_type` in `kwargs`, the payload and confirmation don't agree on `card_type` in the two tests mentioned in the next line.

Pomax:
It might be worth figuring out why the two tests mentioned seem to be relying on a kind of data passing that does not seem to happen during live user interaction for the things the two tests are supposed to be for.

(And this is fairly complex code, so it's entirely possible I misunderstood something, but as far as I can tell at this point, these two tests seem to pass because they do things differently from the real code, which means they might not actually be testing what they're supposed to)

donate/payments/tasks.py
```diff
@@ -62,11 +63,15 @@ def send_transaction_to_basket(data):
...
-            return process_method(notification)
-
-    def process_subscription_charged_successfully(self, notification):
+        kind = notification.kind
+        if kind == 'subscription_charged_successfully':
+            self.subscription_charged_successfully(notification)
+        elif kind == 'subscription_charged_unsuccessfully':
+            self.subscription_charged_unsuccessfully(notification)
+        elif kind == 'dispute_lost':
+            self.dispute_lost(notification)
```
Pomax: this code was a bit _too_ magical. This rewrite changes things so that there are three explicit possible code paths, rather than an unknown number of potential failure paths.

donate/payments/tests/test_views.py
```diff
@@ -489,34 +492,6 @@ def test_get_success_url(self):
             reverse('payments:newsletter_signup')
         )
 
-    def test_get_transaction_details_for_session(self):
```
Pomax: Monthly does not result in a transaction, instead it yields a subscription, with the actual transaction getting triggered via web hook. As such we removed this test.

donate/payments/views.py
```diff
@@ -314,27 +314,35 @@ def process_monthly_transaction(self, form):
...
                 }
             ])
-            return self.success(
-                result,
-                form,
-                transaction_id=result.subscription.id,
-                last_4=payment_method.last_4,
-                send_data_to_basket=False,
-            )
+            # Bypass self.success, as this is not a transaction but a subscription:
```
Pomax: Monthly does not involve a transaction, so falling through to success() which then falls through to getting the transaction details, isn't actually required. We can immediately redirect to the success URL

donate/payments/tasks.py
alanmoo: But then these are redefined in the actual methods below...why not just a comment that says "get the notification kind and call the class method with the matching name"?

donate/payments/tasks.py
```diff
@@ -142,13 +147,14 @@ def process_subscription_charged_successfully(self, notification):
             'transaction_id': last_tx.id,
             'subscription_id': notification.subscription.id,
             'project': custom_fields.get('project', ''),
+            'card_type': 'Unknown' if is_paypal else last_tx.credit_card_details.card_type,
```
alanmoo: At first glance, this seems like it might break when we start accepting other payment types. (I'm guessing they won't all have `credit_card_details`, but I could be wrong.) Can we make it a bit more defensive?

donate/payments/views.py
alanmoo: The more I think about it, this feels like setting us up to missing something in the future. I suspect we should either rename `success` to be something like `transactionSuccess` or check if this was indeed a transaction inside the success method.

I'm trying to think about all the different code pathways we have in this stack and keep them from branching out as much as possible/keep the flow clear.

alanmoo:
(Functionally, this works and makes it to Salesforce; I tested a Visa transaction and a Paypal transaction, the latter of which got "Unknown" as expected.)

donate/payments/tasks.py
Pomax: Because it really was _too_ magical. The refactor was trivial and hardens a part of the code that we needed to touch anyway.

donate/payments/views.py
Pomax: I'm in favour of renaming `.success` to a function that makes it explicit that this is only for transactions.

donate/payments/tasks.py
alanmoo: Ok, you care about this more than I do.

donate/payments/tasks.py
Pomax: I would recommend we do this in followup, then, because the line right below this will also be affected by any "not-paypal, not-creditcard" transaction, and I based this line off of that one.

alanmoo: Fair enough

Pomax:
I've renamed `.success` to `.handle_successful_transaction` so that its role and use is explicit. I recommend we harden the "what if it's not paypal but _also_ not credit cards?" aspects separately, as there is preexisting code that will be affected by that, too (some of which was used as guideline for changes in this PR)

alanmoo:
I tested this again, still works.

Pomax:
Filed https://github.com/mozilla/donate-wagtail/issues/648 as follow-up for the "not-paypal should probably not automatically mean credit card" flag.

donate/payments/tasks.py
cadecairos: I've got no problem using if/elif over `getattr`.

PR #633, opened 2019-11-29
Fix the Stripe charge webhook

### LINKED ISSUE #632: Stripe Webhook: TypeError: the JSON object must be str, bytes or bytearray, not dict
https://sentry.prod.mozaws.net/operations/donate-stage/issues/6761332/

```
TypeError: the JSON object must be str, bytes or bytearray, not dict
  File "rq/worker.py", line 822, in perform_job
    rv = job.perform()
  File "rq/job.py", line 605, in perform
    self._result = self._execute()
  File "rq/job.py", line 611, in _execute
    return self.func(*self.args, **self.kwargs)
  File "donate/payments/tasks.py", line 257, in process_stripe_webhook
    json.loads(form_data),
  File "__init__.py", line 341, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
```

cadecairos:
1. Don't use Form and FormView, since unlike BrainTree, Stripe webhook request bodies are not form encoded.
2. Fix various issues accessing the data about the charge, subscription, invoice, and metadata
3. Update tests for the change in data formatting introduced by the above changes.

Fixes #632

Testing locally is tricky, and requires a reverse-proxy so that you can configure Stripe in test mode to replay webhook events to your local instance of the site. I'd suggest reviewing the code for any glaring issues or omissions, then we can put the endpoint through some more testing in the staging environment.

### DIFF
--- a/donate/payments/stripe_webhooks.py
+++ b/donate/payments/stripe_webhooks.py
@@ -1,25 +1,27 @@
-from django import forms
+import json
 from django.http import HttpResponse, HttpResponseBadRequest
 from django.utils.decorators import method_decorator
 from django.views.decorators.csrf import csrf_exempt
-from django.views.generic import FormView
+from django.views.generic import View
 
 from .tasks import queue, process_stripe_webhook
 
 
-class StripeWebhookForm(forms.Form):
-    data = forms.TextInput()
-
-
 @method_decorator(csrf_exempt, name='dispatch')
-class StripeWebhookView(FormView):
-    form_class = StripeWebhookForm
+class StripeWebhookView(View):
     http_method_names = ['post']
 
-    def form_valid(self, form):
-        signature = self.request.META['HTTP_STRIPE_SIGNATURE']
-        queue.enqueue(process_stripe_webhook, form.cleaned_data, signature=signature)
-        return HttpResponse()
+    def post(self, request):
+        signature = request.META['HTTP_STRIPE_SIGNATURE']
 
-    def form_invalid(self, form):
-        return HttpResponseBadRequest()
+        if not signature:
+            return HttpResponseBadRequest(reason='HTTP_STRIPE_SIGNATURE must be provided')
+
+        try:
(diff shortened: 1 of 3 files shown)


donate/payments/stripe_webhooks.py
```diff
@@ -1,25 +1,27 @@
...
 
 from .tasks import queue, process_stripe_webhook
 
 
-class StripeWebhookForm(forms.Form):
-    data = forms.TextInput()
-
-
 @method_decorator(csrf_exempt, name='dispatch')
-class StripeWebhookView(FormView):
```
cadecairos: Turns out the only reason `FormView` was used for the Braintree webhook endpoint is because Braintree form encodes their webhook payloads. Stripe uses JSON, so the changes here support that.

donate/payments/tasks.py
```diff
@@ -206,27 +205,26 @@ def process_charge_succeeded(self, event):
...
+        if 'thunderbird' in metadata:
+            description = 'Thunderbird monthly'
+        elif 'glassroomnyc' in metadata:
+            description = 'glassroomnyc monthly'
         else:
-            update_data['description'] = 'Mozilla Foundation Monthly Donation'
+            description = 'Mozilla Foundation Monthly Donation'
 
         try:
-            stripe.Charge.modify(charge_id, update_data)
```
cadecairos: This is what happens when you assume the signature of update methods is the same across languages (in this case, stripe-node VS stripe-python

donate/payments/tasks.py
```diff
@@ -248,13 +246,9 @@ def process_webhook(form_data):
 
 
 def process_stripe_webhook(form_data, signature=None):
-    if not signature:
```
cadecairos: moved signature check to the webhook handler

youriwims:
@cadecairos I'm passing this to Lucie since it'll take me too much time to get up to speed and she can most likely help you get it merged much sooner 🚀

PR #662, opened 2019-12-10
Feature/min amounts rebased

Pomax:
This is a rebase of https://github.com/mozilla/donate-wagtail/pull/439, with some code refactored for easier maintenance, and two tests updated because the rebase caused them to break.

Closes #439

### DIFF
--- a/donate/payments/forms.py
+++ b/donate/payments/forms.py
@@ -2,14 +2,16 @@ from django import forms
 from django.conf import settings
 from django.utils.safestring import mark_safe
 from django.utils.translation import gettext_lazy as _
-from django.utils.translation import pgettext_lazy
+from django.utils.translation import get_language, pgettext_lazy
 
 from django_countries.fields import CountryField
 
 from donate.core.utils import is_donation_page
+from donate.core.templatetags.util_tags import format_currency
 from donate.recaptcha.fields import ReCaptchaField
 
 from . import constants
+from .utils import get_currency_info
 
 
 # Global maximum amount value of 10 million, not currency-specific, intended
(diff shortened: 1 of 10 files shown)


donate/payments/forms.py
```diff
@@ -2,22 +2,53 @@
...
+    def clean(self):
+        cleaned_data = super().clean()
+        amount = cleaned_data.get('amount', False)
+        currency = cleaned_data.get('currency', False)
+        if amount and currency:
+            currency_info = get_currency_info(currency)
+            min_amount = currency_info['minAmount']
+            if amount < min_amount:
+                raise forms.ValidationError({
+                    'amount': _('Donations must be %(amount)s or more') % {'amount': format_currency(
```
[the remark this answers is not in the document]
Pomax: That `_` is a shorthand for `gettext`, which is a Django convention that is about as magical as everything else. Especially fun if you have JS devs trying to get into Django.

donate/templates/payment/card_upsell_master.html
```diff
@@ -41,6 +41,10 @@ <h1 class="heading heading--primary heading--bottom-margin">{% trans "Thank you
                             {% render_form_field form.amount %}
                         </div>
 
+                        {% for field in form.hidden_fields %}
```
Pomax: @alanmoo I don't actually know what this does... and whether it's needed or is the template equivalent of an accidental print()

donate/payments/forms.py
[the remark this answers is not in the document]
TheoChevalier: Yup, this should just work. Do you confirm this isn’t the final copy? Just so that I know I shouldn’t extract strings til the final one lands

donate/templates/payment/card_upsell_master.html
alanmoo: This spits out the currency in the upsell and I believe it's important to be able to grab the minimum for that currency.

alanmoo: This injects the currency into the upsell view, which is necessary to be able to determine how much the minimum should be.

Issue #627, opened 2019-11-28
Integrate mozilla-django-oidc for SSO logins


Pomax:
See https://github.com/mozilla/donate-wagtail/issues/616

This issue is for integrating https://github.com/mozilla/mozilla-django-oidc so that SSO can be use to log into the CMS. The developer portal folks recently added this, too, so we should be able to follow their approach.

- docs: https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html
- commit that adds oidc to the dev portal [is here](https://github.com/mdn/developer-portal/commit/96d3c736a730565803d6ef665cbab19c1da74831)
- This will need Moz SSO creds, for which an issue has been filed with mofo devops (722)

We'll want to make sure that users that are created by login _attempts_ are not granted any rights to the CMS. A login attempt should create the user account, but admin action should be required to bestow access rights to those accounts, once created.

Pomax:
@alanmoo do we know if the TB team will need cms-level access, or do we only need SSO for the standard moz oauth0 track?

alanmoo:
It looks like they've got SSO accounts, so we should be good on that front.

Issue #617, opened 2019-11-25
Enable SSO for Donate-wagtail in Production


alanmoo:
At the suggestion of our security team (and because we've been talking about this for a while), we should link the donate stack to SSO. In the future, we'll link the Foundation site as well, but that's outside the scope of this epic.

Pomax:
all tasks have been completed

PR #666, opened 2019-12-12
Adding SSO login capability using Mozilla's OIDC connector

Pomax:
https://donate-wagtail-staging-pr-666.herokuapp.com/admin

image

Following the broad (and some specific) strokes as outlined in https://mozilla-django-oidc.readthedocs.io/en/stable/installation.html and implemented by the dev portal folks over on https://github.com/mdn/developer-portal/commit/96d3c736a730565803d6ef665cbab19c1da74831

Closes https://github.com/mozilla/donate-wagtail/issues/627, part of epic https://github.com/mozilla/donate-wagtail/issues/617

**NOTE** we'll probably want dedicated production creds before we push this from staging to production.

**NOTE 2** we'll need to add everyone's accounts on staging before SSO will work once we land this PR.


README.md
```diff
@@ -139,18 +149,18 @@ Example from donate.mozilla.org:
 
 _Notes_: We want to keep the `trigger_welcome` at `N` and the `format` to `html`. We don't have the country info for now, but from what I understood, it's something we want to change.
 
-### Webhook Configuration
+## Review App
```
Pomax: this is the weirdest diff. I literally moved the webhook config section.

donate/settings.py
```diff
@@ -581,3 +593,45 @@
...
+OIDC_CREATE_USER = False
+
+# How frequently do we check with the provider that the user still exists.
+OIDC_RENEW_ID_TOKEN_EXPIRY_SECONDS = 15 * 60
+
+OIDC_RP_SIGN_ALGO = "RS256"
+OIDC_RP_CLIENT_ID = env("OIDC_RP_CLIENT_ID")
+OIDC_RP_CLIENT_SECRET = env("OIDC_RP_CLIENT_SECRET")
+
+# These values should be overwritten by thunderbird/other instances:
```
Pomax: ... or.... should they?

alanmoo: Can we overwrite the below strings in Thunderbird when they're not being pulled from env vars?

@cadecairos This is going to mean another refactor of django configurations, isn't it. :-|

env.default
```diff
@@ -64,3 +64,8 @@ STRIPE_WEBHOOK_SECRET=
 
 # Sentry
 SENTRY_DSN=
+
+# OIDC
+USE_CONVENTIONAL_AUTH=False
```
alanmoo: Feels like we should set this to True here so devs running this and copying env.default to .env just have things work. (Though that should change once django-configurations lands, if we ever stop getting ahead of it)

README.md
```diff
@@ -139,18 +149,18 @@ Example from donate.mozilla.org:
...
-There's a webhook endpoint for processing Braintree events. The events it supports are:
+Opening a PR will automatically create a Review App in the `donate-wagtail` pipeline. A slack bot posts credentials and links to Review Apps in to the `mofo-ra-donate-wagtail` channel.
 
-* `subscription_charged_successfully`
-* `subscription_charged_unsuccessfully`
-* `dispute_lost`
+This only work for Mo-Fo staff: you will need to manually open a Review App on Heroku for PRs opened by external contributors.
 
-The endpoint accepts requests on `/braintree/webhook/` and will verify the payload signature to ensure it's a legitimate event. [Documentation for Braintree webhooks can be found here](https://developers.braintreepayments.com/guides/webhooks/overview).
+## SSO and admin logins
```
alanmoo: Do we need to note who to go to for new Auth0 accounts? Is it "File a Bugzilla ticket"?

donate/settings.py
Pomax: no idea, but hey let's start even earlier than that: _do thunderbird folks need access or not_ because that's been an unanswered question for like three weeks now. If not, then we literally don't care >_>

Pomax:
<< Do we actually need a non-SSO login option, or does that defeat the point of this?  >>

It, in fact, violates the security around SSO. So no, we should absolutely not have a standard login fallback.

env.default
Pomax:
```suggestion
USE_CONVENTIONAL_AUTH=True
```
Good point

README.md
Pomax: It is, but that bugzilla ticket would then come back to us going "this person is requesting access, is that okay?" so I'd recommend that for this we go with "ask us", and then if we agree those folks really need access, we file the bugzilla issue to request they get added to our donate-wagtail ldap group.

@cadecairos thoughts?

alanmoo: Right, but how do _we_ remember that we need to file a Bugzilla ticket? 😄

Pomax:
conflicts? what conflicts >_>

README.md
Pomax: is there a designated "code owner" that we can say people should poke? (filing a bugzilla issue is effectively a given, since only moco has LDAP super powers)

cadecairos: To be honest, In the interest of avoiding spear phishing attempts here, I'd rather not make the process "email somebody to ask for access"

Pomax: I was thinking "contact them on slack", but that's still a fair point.

Pomax:
@alanmoo @cade I've updated the README.md to explain "how to SSO" for local dev and what the new user policy is. It feels overkill to add "how to get SSO creds" to the README.md, though: anyone who actually needs to work on this code is more than likely to have been instructed to do so by us, which means we'll be able to explain how to get working creds?

cade:
But... how do _**I**_ get SSO creds?

alanmoo:
Oh, we'll just file a Bugzilla ticket for you. At this point you're basically increasing our bus factor anyway 😜

PR #677, opened 2019-12-19
Fetch environment variables over XHR

cadecairos:
This will ensure that pipeline promotions work again.


cadecairos:
This strategy is modeled after the one we use on the foundation site.

source/js/main.js
```diff
@@ -11,6 +11,17 @@ import CopyURL from "./components/copy-url";
 import Accordion from "./components/accordion";
 import "./components/newsletter";
 
+function fetchEnv(callback) {
+  let envReq = new XMLHttpRequest();
+
+  envReq.addEventListener("load", () => {
+    callback.call(this, JSON.parse(envReq.response));
```
Pomax: `JSON.parse` can throw, and the callback should not need any execution context rebinding; it does not rely on a `this` keyword for binding the Sentry values.

```suggestion
    let data = {};
    try {
      data = JSON.parse(envReq.response);
    } catch (e) {
      // we could report this error, or we could just leave it be
    }
    callback(data);
```

cadecairos: I'm okay with throwing the parse in a try/catch, but if it ever did pass invalid json, that'd mean `django.http.JsonResponse` is severely broken. If we're genuinely concerned here we should also make this change on the foundation site.

cadecairos: I don't see a need to report an error here to Sentry since I have doubts that django would break their JsonResponse formatter.

Issue #692, opened 2020-01-06
Conventional login no longer works


Pomax:
When running with `USE_CONVENTIONAL_AUTH=True` the username/password login system no longer works. The reason here is that it will always load the OIDC auth backend, which should only happen if `USE_CONVENTIONAL_AUTH` is False.

PR #693, opened 2020-01-06
don't load OIDC backend when using conventional login

Pomax:
closes https://github.com/mozilla/donate-wagtail/issues/692


Pomax:
superceded.

PR #705, opened 2020-01-07
django conf. optimization for shared inheritance

Pomax:
This unifies the three runtime environments (dev/staging/prod) so that new setting objects (e.g. Salesforce) just need to be added in a single place.

I'm not intimately familiar with the inheritance ordering that D.C. requires though, so the argument ordering might be incorrect for some of these.

### DIFF
--- a/donate/settings/shared_settings.py
+++ b/donate/settings/shared_settings.py
@@ -1,0 +1,36 @@
+from configurations import Configuration
+from .base import Base
+from .oidc import OIDC
+from .database import Database
+from .redis import Redis
+from .s3 import S3
+from .secure import Secure
+from .sentry import Sentry
+from .braintree import Braintree
+
+class SharedSettings(Base, OIDC, Database, Redis, S3, Braintree, Configuration):
+    @classmethod
+    def pre_setup(cls):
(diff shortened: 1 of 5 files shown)


donate/settings/shared_settings.py
```diff
@@ -0,0 +1,36 @@
...
+    @classmethod
+    def setup(cls):
+        super().setup()
+
+    @classmethod
+    def post_setup(cls):
+        super().post_setup()
+
+
+class SharedSettingsDeployment(Secure, Sentry, SharedSettings):
```
Pomax:
```suggestion
class SharedSettingsDeployment(SharedSettings, Secure, Sentry):
```

donate/settings/shared_settings.py
```diff
@@ -0,0 +1,36 @@
...
+    def setup(cls):
+        super().setup()
+
+    @classmethod
+    def post_setup(cls):
+        super().post_setup()
+
+
+class SharedSettingsDeployment(SharedSettings, Secure, Sentry):
+    @classmethod
```
alanmoo: What's the benefit of  having these classmethods repeated?

PR #713, opened 2020-01-09
Add in the option to do a newsletter signup using a POST url after donating

Pomax:
This will be useful for TB and special campaigns.


donate/payments/views.py
```diff
@@ -760,8 +762,16 @@ def get_initial(self):
...
-        if send_data_to_basket:
-            data = form.cleaned_data.copy()
+        data = form.cleaned_data.copy()
+
+        if hasattr(settings, 'THUNDERBIRD'):
+            newsletter_url = settings.POST_DONATE_NEWSLETTER_URL
+            data = parse.urlencode({'EMAIL': data['email']}).encode()
+            req =  request.Request(newsletter_url, data=data)
+            res = request.urlopen(req)
+            # Do we want Sentry logging for res.status here, in case it's not 200?
```
Pomax: @cadecairos what's the right way to do this, such that the data goes to the folks that need to read it?

donate/settings/__init__.py
```diff
@@ -7,7 +7,7 @@
 from .sentry import Sentry
 from .testing import Testing
 from .development import Development, ThunderbirdDevelopment
-from .review_app import ReviewApp
+from .review_app import ReviewApp, ThunderbirdReviewApp
```
Pomax: same as https://github.com/mozilla/donate-wagtail/pull/719 but necessary here in order to test things.

cadecairos:
In localdev, with the POST URL undefined in my `.env` I hit this server error:

image

donate/payments/views.py
```diff
@@ -760,8 +761,23 @@ def get_initial(self):
         return initial
 
     def form_valid(self, form, send_data_to_basket=True):
-        if send_data_to_basket:
-            data = form.cleaned_data.copy()
+        data = form.cleaned_data.copy()
+
+        if hasattr(settings, 'POST_DONATE_NEWSLETTER_URL'):
```
cadecairos: since it defaults to `None`, we need to check for that, something like:
```suggestion
        if hasattr(settings, 'POST_DONATE_NEWSLETTER_URL') and settings.POST_DONATE_NEWSLETTER_URL:
```

cadecairos: actually, since we use django-environ to default it to None, we don'te really need a `hasattr` check right? it will either be a `String` or `None`

cadecairos: updated the suggestion.

Pomax:
@cade applied that suggestion, good thinking.

cade:
> @cade applied that suggestion, good thinking.

It was nothing, really.

I mean, literally, _nothing_.

Pomax:
I need a browser extension that just goes "no, Pomax, you do not mean @cade, really".

cade:
But then I would be robbed of the joy of trolling your PRs! I hope you do not make said extension.

alanmoo:
@cade I'm leaving Mozilla at the end of the month and (don't tell the others) I think I'm going to miss you most of all.

donate/thunderbird/templates/fragments/donate_form_disclaimer.html
```diff
@@ -9,7 +9,7 @@
 {% endblock %}
 
 {% block faq_contact %}
-    {% blocktrans with faq_url='/faq' email='thunderbird-donate@mozilla.org' trimmed %}
+    {% blocktrans with faq_url='/faq' email='give@thunderbird.net' trimmed %}
```
alanmoo: Is this a real email address? @stephaniemcv?

stephaniemcv: Those are both real email addresses.

donate/thunderbird/templates/payment/newsletter_signup.html
```diff
@@ -0,0 +1,10 @@
+{% extends "payment/newsletter_signup_master.html" %}
+
+{% load i18n %}
+
+{% block thank-you %}
+    <h1 class="heading heading--primary heading--bottom-margin">{% trans "Thank you for your generous gift" %}</h1>
+    <p>{% blocktrans %}
```
TheoChevalier: And here too, also because it will then match an existing string and use that translation

```suggestion
    <p>{% blocktrans trimmed %}
```

Issue #665, opened 2019-12-10
improve UX when using a custom amount + paypal button


Pomax:
Currently a "too low" amount in the custom amount textbox followed by the paypal button leads to really poor UX (popping up the paypal dialog, which then closes again, after which the main site shows an error for the textbox with a mathematical error to "please use a value greater or equal than 2").

After talking with @beccaklam this should be changed to something where we intercept the paypal click and instead immediately show the error without ever hitting the paypal flow (either by having a click listener on the iframe element itself, or an overlay div), or allowing the click through if the amount fllled in is fine.

Pomax:
After some digging, and the restrictions on `<iframe>` in general, this is going to have to be an overlay div with click interception, only allowing the click through to `overlaydiv.parentElement.querySelector('iframe')` when there are no form validation errors.

PR #717, opened 2020-01-09
overlay a click-interceptor for paypal when the amount chosen is too low

Pomax:
closes https://github.com/mozilla/donate-wagtail/issues/665

Note that for review purposes, the overlay is red because it has a solid background color set so you can see the overlay existing at all. Prior to landing, that background color will get removed, so no one will ever see it.

### DIFF
--- a/source/sass/abstracts/_variables.scss
+++ b/source/sass/abstracts/_variables.scss
@@ -79,7 +79,8 @@ $breakpoints: (
 */
 
 $z-index: (
-  1: 10,
+  hidden: -1,
+  base: 1,
   to-do: 20,
   search: 30,
   nav: 40,
@@ -87,8 +88,8 @@ $z-index: (
   nav-desktop: 60,
   ui-overlay: 70,
   sticky-message: 80,
-  9: 90,
-  modal: 100
+  modal: 100,
+  overlay: 200
 );
 
 // --------------------------------- Grid Dimensions --------------------------------------
--- a/source/sass/components/_payments.scss
+++ b/source/sass/components/_payments.scss
@@ -32,7 +32,6 @@
         right: 0;
         bottom: 0;
         left: 0;
-        background: $color--warning;
         cursor: pointer;
       }
     }
--- a/source/sass/components/_payments.scss
+++ b/source/sass/components/_payments.scss
@@ -19,9 +19,21 @@
     }
 
     &--paypal {
+      position: relative;
       border-radius: 4px;
       .paypal-disabled & {
         display: none;
+      }
+      &--overlay {
+        display: none;
+        position: absolute;
+        @include z-index(overlay);
+        top: 0;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        background: $color--warning;
+        cursor: pointer;
       }
     }
   }
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -13,48 +13,25 @@ let rules = [
   }
 ];
 
-let main = {
-  entry: `./source/js/main.js`,
-  output: {
-    path: frontendPath,
-    filename: `main.compiled.js`
-  },
-  module: {
-    rules
-  }
-};
+function generate(name) {
+  return {
+    entry: `./source/js/${name}.js`,
+    output: {
+      path: frontendPath,
+      filename: `${name}.compiled.js`
+    },
+    module: {
+      rules
+    },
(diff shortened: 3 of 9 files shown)


source/sass/abstracts/_variables.scss
```diff
@@ -79,16 +79,17 @@ $breakpoints: (
 */
 
 $z-index: (
-  1: 10,
```
Pomax: there is no scss that uses `@include zindex(1)`

source/sass/abstracts/_variables.scss
```diff
@@ -79,16 +79,17 @@ $breakpoints: (
...
+  hidden: -1,
+  base: 1,
   to-do: 20,
   search: 30,
   nav: 40,
   header: 50,
   nav-desktop: 60,
   ui-overlay: 70,
   sticky-message: 80,
-  9: 90,
```
Pomax: there is no scss that uses `@include zindex(9)`

source/sass/components/_payments.scss
```diff
@@ -19,10 +19,22 @@
...
       }
+      &--overlay {
+        display: none;
+        position: absolute;
+        @include z-index(overlay);
+        top: 0;
+        right: 0;
+        bottom: 0;
+        left: 0;
+        background: $color--warning;
```
Pomax: this line will be removed once this is R+

webpack.config.js
```diff
@@ -13,48 +13,25 @@ let rules = [
...
+  return {
+    entry: `./source/js/${name}.js`,
+    output: {
+      path: frontendPath,
+      filename: `${name}.compiled.js`
+    },
+    module: {
+      rules
+    },
+    devtool: `none` // see https://webpack.js.org/configuration/devtool/
```
Pomax: The `devtool` instruction was missing, which means it defaulted to creating a source map based on `eval`, which made _everything_ break.

Fuck magical implicit webpack nonsense, it absolutely destroys a day's worth of productivity.

beccaklam:
Hey @Pomax so the changes are looking good but I have some questions:
1) What is the exact behaviour we're going for here? Should the comment box ("Please select a value ...") show up on every screen? It's working for me in Chrome for all screen widths, only desktop for Firefox (though I'm not sure if this is because of the Inspector, seems to work fine on Android on Browser Stack), and is very buggy is Safari. 
2) What is the red overlay you're referring to? IS it the red outline around the box?
Screen Shot 2020-01-13 at 3 52 29 PM

Pomax:
@beccaklam you should see a red overlay on the paypal button when the custom amount is too low:

image

Also it turns out we don't control that message: it's default browser form validation behaviour, which on the one hand means it might not be a super nice error message, but on the other hand does mean that we get proper, vetted localization for free, so we probably don't want to fix that message in this PR (if at all).

beccaklam:
> @beccaklam you should see a red overlay on the paypal button when the custom amount is too low:
> 
> image
> 
> Also it turns out we don't control that message: it's default browser form validation behaviour, which on the one hand means it might not be a super nice error message, but on the other hand does mean that we get proper, vetted localization for free, so we probably don't want to fix that message in this PR (if at all).

Thanks for the clarification, @Pomax. The functionality looks good, though I do agree with @alanmoo 's comment above 👆 And one design note: would it be possible to have the red outline go all the way around the box, including the '$' sign portion? It looks a bit odd cut off. If it's too big of a lift, that's okay we can still ship it.

Screen Shot 2020-01-14 at 10 29 30 AM

alanmoo:
@beccaklam How's an inset error state instead?
image

```
#one-time-amount-other-input:invalid {
  box-shadow: red inset 0px 0px 1.5px 1px;
}
```

beccaklam:
> @beccaklam How's an inset error state instead?
> image
> 
> ```
> #one-time-amount-other-input:invalid {
>   box-shadow: red inset 0px 0px 1.5px 1px;
> }
> ```

Oh that's much better. Thanks @alanmoo !

Pomax:
The problem is mostly one of timing: we're listening for `input` and `change` events at the moment, because if we only listened to just `change` then the user can type a value and click the paypal button without the form every having "changed" first.

I'm not sure we can get around that without a lot more code and introducing some edge cases =(

Pomax:
@alanmoo @beccaklam inset highlight added, but let's leave the timing issue as something separate.

Pomax:
nice, then I'll remove the background color in preparation for landing this.

Issue #716, opened 2020-01-09
custom value can have a value while showing as empty


Pomax:
STR
1. click custom field, fill in '2'
2. click one of the predefined amount buttons
3. click the currency label _next to_ to the custom value input

The custom value now shows as empty but the _actual_ stored value is still 2, and you can initial payment even though no value is visible.

This doesn't feel like a blocker, but I tried to see if there was an easy fix and right now there isn't, as far as I can tell.

ragnarokatz:
@Pomax hello, I would like to work on this issue please.

ragnarokatz:
@Pomax I have created a PR to address this issue. Please take a look at it when you have time, thanks.

PR #721, opened 2020-01-11
Fix issue 716: custom value can have a value while showing as empty

ragnarokatz:
This PR fixes issue #716 

After a bit of console logging, I've traced to the source of the problem:

In the original event listener defined for when clicking on preset donation amounts, only the values for html elements containing tag `data-other-amount` were cleared. However, the actual custom donation amount being submitted through the form is stored in the html element containing tag `data-other-amount-radio`, which was not cleared.

I modified the event listener for clicking on preset donation amounts to clear values stored in html elements with tag `data-other-amount-radio` as well.

### DIFF
--- a/source/js/components/currency-selector.js
+++ b/source/js/components/currency-selector.js
@@ -202,10 +202,16 @@ class CurrencySelect {
     const presets = document.querySelectorAll(
       ".donation-amount__radio:not([data-other-amount-radio])"
     );
+
+    const customAmounts = document.querySelectorAll(
+      "[data-other-amount-radio]"
+    );
+
     presets.forEach(input => {
       input.addEventListener("change", _ => {
         if (input.checked) {
           otherAmountInputs.forEach(other => (other.value = ""));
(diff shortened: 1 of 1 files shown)


source/js/components/currency-selector.js
```diff
@@ -205,7 +205,13 @@ class CurrencySelect {
     presets.forEach(input => {
       input.addEventListener("change", _ => {
         if (input.checked) {
-          otherAmountInputs.forEach(other => (other.value = ""));
+          otherAmountInputs.forEach(other => {
+            other.value = "";
+            var radio = other.parentNode.querySelector(
```
Pomax: let's move this up to outside of the `presets.forEach` on line 205, because it's a fixed HTML element, so we don't need to query for it on every iteration. Also, let's call it something indicative of its role, like `customAmount`.

ragnarokatz: fixed.

Pomax:
awesome, thank you for the help!

PR #732, opened 2020-01-13
Custom OG image and page title for Thunderbird

TheoChevalier:
This part remained to be customized


TheoChevalier:
Sorry @Pomax, forgot to update the actual page title in `base.html`

Pomax:
ah, good catch. R+

PR #733, opened 2020-01-13
Fix issue 716 custom value can have value while empty

### LINKED ISSUE #716: custom value can have a value while showing as empty

Pomax:
local branch version of https://github.com/mozilla/donate-wagtail/pull/721, for review app purposes

closes #716

### DIFF
--- a/source/js/components/currency-selector.js
+++ b/source/js/components/currency-selector.js
@@ -202,10 +202,16 @@ class CurrencySelect {
     const presets = document.querySelectorAll(
       ".donation-amount__radio:not([data-other-amount-radio])"
     );
+
+    const customAmounts = document.querySelectorAll(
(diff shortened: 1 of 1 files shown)


Pomax:
This still allows the CC/Paypal button to be clicked without an amount filled in, so this might need a little bit more work.

Pomax:
We might need to add the `required` attribute to the input field when focus is given to it or the label, and remove it when any of the predefined buttons are pressed.

PR #739, opened 2020-01-14
Implement a Redirect View for Thunderbird that preserves query parameters

### LINKED ISSUE #738: Redirect Thunderbird donate landing page
We need an optional redirect handler for `/thunderbird/` that preserves query params so that we can direct `https://donate.mozilla.org/thunderbird/` to the new thunderbird donation stack.

Use case: The Thunderbird site redirects people to `donate.mozilla.org/thunderbird?tbdownload=true` after downloading. Once we move Thunderbird's donate page to `give.thunderbird.net`, we still want to know if a user is coming from a download to show them the banner. Thus, we want to preserve the query params; this is only needed temporarily until the Thunderbird site has been updated, but doing so frees us from being strictly dependent on that.

cadecairos:
Fixes #738


donate/views.py
```diff
@@ -11,3 +11,13 @@ class EnvVariablesView(View):
...
         return JsonResponse(settings.FRONTEND)
+
+
+class ThunderbirdRedirectView(RedirectView):
+    """
+    A view that redirects requests to give.thunderbird.net, preserving query params
+    """
+
+    url = 'https://give.thunderbird.net/'
+    permanent = False # True
```
cadecairos: got ahead of myself. Should we leave this as non-permanent so that query params are properly appended to the redirect if visited more than once on a machine?

donate/settings/development.py
```diff
@@ -40,6 +40,7 @@ def post_setup(cls):
 
 class ThunderbirdDevelopment(Development, ThunderbirdOverrides, Configuration):
     INSTALLED_APPS = ThunderbirdOverrides.INSTALLED_APPS + Development.INSTALLED_APPS
+    ENABLE_THUNDERBIRD_REDIRECT = False
```
Pomax: do we need this if we're already setting it to `false`  by default in `environment.py`?

Pomax: (also, with every configuration explicitly set to `False`, how would one set it to `True`?)

Pomax: Looking at this again... shouldn't these all be `True` in the TB configurations?

PR #753, opened 2020-01-16
Add context to dropdown strings / prevent string removal

TheoChevalier:


donate/templates/pages/core/contributor_support_page_master.html
```diff
@@ -50,14 +50,14 @@ <h1 class="heading heading--primary">
                         <label for="00N0B000006X3fw">{% trans "Language" %}</label>
                         <div class="form-item__wrapper form-custom-select">
                             <select id="00N0B000006X3fw" name="00N0B000006X3fw" title="Language">
-                                <option value="">--{% trans "None" %}--</option>
+                                <option value="">--{% trans "None" context "Option in “Language” dropdown" %}--</option>
```
alanmoo: The quote situation is confusing here- can we use single quote instead of fancy doubles?

TheoChevalier: sure!

PR #773, opened 2020-01-21
Custom Ways to give template for Thunderbird

TheoChevalier:
Customizing the page with what we had on the old stack


donate/thunderbird/templates/pages/core/ways_to_give_page.html
```diff
@@ -0,0 +1,25 @@
+{% extends "pages/core/ways_to_give_page_master.html" %}
+
+{% load i18n %}
+
+{% block tb_memo_eur %}<br><b>{% trans "Memo field:" %}</b> Thunderbird{% endblock %}
+{% block tb_memo_gbp %}<br><b>{% trans "Memo field:" %}</b> Thunderbird{% endblock %}
+
+{% block check_payee %}
+    {% blocktrans with thunderbird='Thunderbird' payee='Mozilla Foundation' trimmed %}
+    Please send a check <strong>with a separate note marked “{{thunderbird}}”</strong>, and payable (payee) to “{{payee}}”.
```
alanmoo: You lost some copy here. Intentional?

TheoChevalier: I’ve replicated what we had, so that was intentional, but I’m not 100% sure we don’t actually need that mention for TB too. Checking with Will

Also this makes me realize this line is supposed to say `Please send a check with description/memo line marked “Thunderbird”`.

image

TheoChevalier: Landed some copy edits after chatting with Will. We’re good copy-wise

PR #778, opened 2020-01-22
Migrate Stripe subscriptions to Braintree

cadecairos:
This PR adds an extra step on the charge.succeeded webhook endpoint that will cancel a subscription if one exists, and recreates it on Braintree.


cadecairos:
proposal: Let's do some integration testing on Staging, since this is behind a feature flag (off by default)

donate/payments/tasks.py
```diff
@@ -245,6 +252,82 @@ def process_charge_succeeded(self, event):
...
+            return
+
+        # Grab the Payment Method token that represents the customers payment information in the Braintree Vault
+        payment_method_token = BT_customer.payment_methods[0].token
+
+        # Determine the Braintree plan, merchant account, and price for the subscription
+        # based on the Stripe subscription
+        BT_plan_id = get_plan_id(subscription.plan.currency)
+        BT_merchant_account = get_merchant_account_id_for_card(subscription.plan.currency)
+        BT_price = zero_decimal_currency_fix(charge.amount, subscription.plan.currency)
```
cadecairos: Note: we 100% want to validate that zero-decimal currency subscriptions properly migrate.

donate/payments/tasks.py
```diff
@@ -245,6 +252,82 @@ def process_charge_succeeded(self, event):
...
+        )
+
+        # Ensure we found a customer record, if not, report it to Sentry as an error
+        if len(BT_customers) == 0:
+            logger.error(f'No Customer record exists in Braintree '
+                         f'for the Stripe customer attached to this charge object:'
+                         f'{charge.id}', exc_info=True)
+            return
+
+        # There should be a single Customer/Card combo on Braintree
```
Pomax: Should we also add an `if(len(BT_customers) > 1):` here, with a sentry warning?

cadecairos: Added in a check and report for this.

donate/payments/tasks.py
```diff
@@ -331,6 +341,71 @@ def process_charge_failed(event):
...
+                      f' Braintree for this charge: {charge.id}'
+            print(message)
+            logger.error(message, exc_info=True)
+            return
+
+        # Determine the Braintree plan, merchant account, and price
+        # for the subscription based on the subscription's attributes
+        BT_plan_id = get_plan_id(charge.currency)
+        BT_merchant_account = get_merchant_account_id_for_card(charge.currency)
+        BT_price = Decimal(zero_decimal_currency_fix(charge.amount, charge.currency))
```
Pomax: yes. very smart.

donate/payments/tasks.py
```diff
@@ -331,6 +341,71 @@ def process_charge_failed(event):
...
+        BT_first_billing_date = datetime.utcfromtimestamp(subscription.current_period_end)
+
+        # Create the Braintree subscription and record the result for analysis
+        BT_create_subscription_result = gateway.subscription.create({
+            'payment_method_token': payment_method.token,
```
Pomax: Do we need to verify that `payment_method.token` exists, where we're checking if `payment_method` exists? Or does this exist by definition and we don't need to worry about this prop possibly not existing?

cadecairos: Token must exist of there is a payment method object. 👍

PR #779, opened 2020-01-23
Remove thunderbird footer signup

alanmoo:
I think this will address a large amount of sentry errors, though I'll also file an issue to re-implement this except with the right newsletter in place.

### DIFF
--- a/donate/templates/pages/base_page_master.html
+++ b/donate/templates/pages/base_page_master.html
@@ -64,34 +64,36 @@
 <div class="footer">
     <div class="footer__container layout">
         <div class="footer__primary-content layout__container">
-            <div class="footer__newsletter newsletter-signup layout__half-col">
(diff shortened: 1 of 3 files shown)


TheoChevalier:
@alanmoo  Looks like this is causing a regression for the top nav menu.
You can’t open the menu anymore on https://thunderbird-remove-thun-28jbgs.herokuapp.com/en-US/

And I see this in the console `21:08:15,832 TypeError: can't access property "addEventListener", t is null main.compiled.5ef6de1600c4.js:16:181550
`

TheoChevalier:
I think this line requires an element that this PR removes https://github.com/mozilla/donate-wagtail/blob/master/source/js/components/newsletter.js#L128

alanmoo:
Fiiiine.

PR #781, opened 2020-01-24
Separate GA ids for thunderbird and foundation

gideonthomas:
This is so that we can setup different GA tracking IDs for MZLA and the foundation.

### DIFF
--- a/source/js/main.js
+++ b/source/js/main.js
@@ -34,6 +34,12 @@ function closeMenu() {
 }
 
 document.addEventListener("DOMContentLoaded", function() {
+  const gaMeta = document.querySelector(`meta[name="ga-identifier"]`);
+  if (gaMeta) {
+    let gaIdentifier = gaMeta.getAttribute(`content`);
+    initializeGA(gaIdentifier);
+  }
+
   // Initialize Sentry error reporting
 
   fetchEnv(envData => {
@@ -88,7 +94,7 @@ document.addEventListener("DOMContentLoaded", function() {
 });
 
 // Google Analytics
-(function() {
+function initializeGA(trackingId) {
   var doNotTrack =
     navigator.doNotTrack || navigator.msDoNotTrack || window.doNotTrack;
   if (!doNotTrack || doNotTrack === "no" || doNotTrack === "unspecified") {
@@ -113,7 +119,7 @@ document.addEventListener("DOMContentLoaded", function() {
     );
 
     if (typeof ga === "function") {
-      ga("create", "UA-49796218-32", "auto");
+      ga("create", trackingId, "auto");
 
       // Ensure we don't pass the email query param to Google Analytics
       var loc = window.location,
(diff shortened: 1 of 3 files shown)


source/js/main.js
```diff
@@ -49,19 +52,69 @@ function closeMenu() {
   tabIndexer();
 }
 
+// Google Analytics
+function initializeGA(trackingId) {
```
gideonthomas: all of this logic is the same as before (taken from GA directly I think).

donate/settings/base.py
```diff
@@ -161,7 +161,8 @@ def MIDDLEWARE(self):
     FRONTEND = {
         'RELEASE_VERSION': env('HEROKU_RELEASE_VERSION'),
         'SENTRY_DSN': env('SENTRY_DSN'),
-        'SENTRY_ENVIRONMENT': env('SENTRY_ENVIRONMENT')
+        'SENTRY_ENVIRONMENT': env('SENTRY_ENVIRONMENT'),
+        'GA_TRACKING_ID': env('GA_TRACKING_ID'),
```
alanmoo: Does this need to be another env var that's specified in all the heroku instances, or can we bake it into the Staging/Production django configuration, with an override for the Thunderbird app?

gideonthomas: it's probably easier to manage as an env var on heroku instances. This does not need to be specified on ALL instances though. If it exists, it uses it, if not, it doesn't.

source/js/main.js
```diff
@@ -49,19 +54,69 @@ function closeMenu() {
...
-      return;
+    const {
+      GA_TRACKING_ID: trackingId,
+      SENTRY_DSN: dsn,
+      RELEASE_VERSION: release,
+      SENTRY_ENVIRONMENT: environment
+    } = envData;
+
+    if (allowGA && typeof ga === "function" && trackingId) {
+      initializeGA(trackingId);
```
cadecairos: Can we guarantee that the GA scripts will be loaded and ready to initialize before this section of code? Seems there's a bit of a race condition at first glance?

gideonthomas: if it doesn't, this code won't run -- hence the `typeof ga === "function"` check

cadecairos: Right, but what if it should run, but doesn't, because the ga script tag injected below hasn't loaded yet? 

In that case, we're losing out on analytics we would have otherwise been able to use.

cadecairos: My concern is that `DOMContentLoaded` is tied to the loading of the inital HTML document, so I don't think it's going to wait for that GA script tag.

gideonthomas: any suggestions on how to fix that?

cadecairos: We can rearrange things a bit and hook  the initialization code into the scripts onload event? We might need to poll for `ga` though..

gideonthomas: yeah so onload doesn't guarantee that the script has run, just that it has loaded. Are you sure you want to introduce `setInterval` or `setTimeout` into this?

PR #783, opened 2020-01-27
Make salesforce help forms work

alanmoo:
This isn't pretty, but it works. I'm open to alternative approaches...once we launch this, but I was moving in the direction of the code we already had.

Steph has confirmed that this makes cases appear in the right place.

```
SALESFORCE_ORGID=00DU0000000IrgO
SALESFORCE_CASE_RECORD_TYPE_ID=0124O000000tDBO
CSP_FORM_ACTION=https://webto.salesforce.com/
SALESFORCE_FORM_URL=https://webto.salesforce.com/servlet/servlet.WebToCase?encoding=UTF-8
```

(Some of these are baked into the production djangoconfig)

### DIFF
--- a/donate/settings/production.py
+++ b/donate/settings/production.py
@@ -13,6 +13,7 @@ from .thunderbird import ThunderbirdOverrides
 
 class Production(Base, Secure, OIDC, Database, Redis, S3, Salesforce, Braintree, Sentry, Configuration):
     DEBUG = False
+    SALESFORCE_FORM_URL = "https://webto.salesforce.com/servlet/servlet.WebToCase?encoding=UTF-8"
 
     @classmethod
     def pre_setup(cls):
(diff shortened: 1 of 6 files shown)


donate/settings/production.py
```diff
@@ -13,6 +13,7 @@
 
 class Production(Base, Secure, OIDC, Database, Redis, S3, Salesforce, Braintree, Sentry, Configuration):
     DEBUG = False
+    SALESFORCE_FORM_URL = "https://webto.salesforce.com/servlet/servlet.WebToCase?encoding=UTF-8"
```
alanmoo: This will likely be used on the MoFo donations too, so I put it in Production.

donate/settings/production.py
```diff
@@ -33,6 +34,7 @@ def post_setup(cls):
 class ThunderbirdProduction(Production, ThunderbirdOverrides, Configuration):
     INSTALLED_APPS = ThunderbirdOverrides.INSTALLED_APPS + Production.INSTALLED_APPS
     ENABLE_THUNDERBIRD_REDIRECT = False
+    SALESFORCE_CASE_RECORD_TYPE_ID = "0124O000000tDBO"
```
alanmoo: Unique to Thunderbird production

Issue #777, opened 2020-01-22
Responsive image bug on landing page


alanmoo:
STR
1. Visit https://give.thunderbird.net/en-US/
2. Zoom out in your browser (make the page smaller). At some point the wordmark for Thunderbird will change to a different version that is too cropped.
Actual results: 
<img width="353" alt="image" src="https://user-images.githubusercontent.com/1682681/72923423-07a7ba00-3d1d-11ea-801b-6941eed8fd25.png">

Expected results:
<img width="615" alt="image" src="https://user-images.githubusercontent.com/1682681/72923434-11312200-3d1d-11ea-864e-0ef9d66b9bab.png">

I suspect this has to do with a combination of the srcset tag and Wagtali behavior, and manifests on a certain configuration of HiDPI displays in regular use (not just zoomed out), so it's worth investigating.

alanmoo:
@nicklee Any chance the problem here has an obvious solution that jumps out at you?

nicklee:
Hi @alanmoo, looks like as you've mentioned this is a mix of wagtail's focus point and the srcset that was implemented:

```
{% image page.featured_image fill-290x95 as imageMobile %}
{% image page.featured_image fill-690x183 as imageTablet %}
{% image page.featured_image fill-350x197 as imageDesktop %}
```

Seems the tablet version is working great:
https://assets.mofoprod.net/thunderbird/images/thunderbird-logo-wordmark-small.2e16d0ba.fill-690x183.png

but the mobile and desktop versions are both awkwardly cropped:

https://assets.mofoprod.net/thunderbird/images/thunderbird-logo-wordmark-small.2e16d0ba.fill-350x197.png

https://assets.mofoprod.net/thunderbird/images/thunderbird-logo-wordmark-small.2e16d0ba.fill-290x95.png

I'm guessing we hadn't considered that a logo that requires exact dimension would be used here, we'd only tested the banners that were included in the designs (these varied in ratio depending on the breakpoint).

It might be worth adding a flag on the landing page template for when a logo is used "Is the image a logo", which just outputs one image that's scaled accordingly for different breakpoints?

PR #798, opened 2020-01-30
Fix thunderbird homepage image

alanmoo:
Band-aid for #777 

Since the image in the CMS is the wordmark.


alanmoo:
Why yes, I will land this after 5PM on my penultimate day with no review. What could go wrong?

PR #823, opened 2020-02-11
Resize but not crop the image shown on landing page

### LINKED ISSUE #777: Responsive image bug on landing page

mmmavis:
Closes #777

To test, resize your browser window & zoom in/out to make sure image shows up well.

Note: The is [the original image](https://assets.mofoprod.net/network/images/highlights/fellowships-that-empower-leaders_1522788145.jpg) I uploaded to the donate review app.

### DIFF
--- a/donate/templates/pages/core/landing_page_master.html
+++ b/donate/templates/pages/core/landing_page_master.html
@@ -9,9 +9,9 @@
             <div class="image-feature column-spacing">
                 {% block featured_image %}
                     <div class="image-feature__container">
-                        {% image page.featured_image fill-290x95 as imageMobile %}
-                        {% image page.featured_image fill-690x183 as imageTablet %}
-                        {% image page.featured_image fill-350x197 as imageDesktop %}
+                        {% image page.featured_image width-290 as imageMobile %}
+                        {% image page.featured_image width-690 as imageTablet %}
+                        {% image page.featured_image width-350 as imageDesktop %}
                         <img srcset="{{ imageMobile.url }} 290w,
                                     {{ imageTablet.url }} 690w,
                                     {{ imageDesktop.url }} 350w"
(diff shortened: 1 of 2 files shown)


donate/templates/pages/core/landing_page_master.html
```diff
@@ -9,9 +9,9 @@
             <div class="image-feature column-spacing">
                 {% block featured_image %}
                     <div class="image-feature__container">
-                        {% image page.featured_image fill-290x95 as imageMobile %}
-                        {% image page.featured_image fill-690x183 as imageTablet %}
-                        {% image page.featured_image fill-350x197 as imageDesktop %}
+                        {% image page.featured_image width-290 as imageMobile %}
```
mmmavis: `fill` resizes AND crops the image. Since we don't have control over what image user uploads and we don't want it to end up being cropped weirdly, I replaced `fill` with `width` to make sure the only treatment we apply is resizing.

Pomax: But wagtail has a "focal area" option for images, that lets people indicate which parts of an image "must" stay in view. Does that not work here?

image

mmmavis: I think the problem is we can’t predict what image CMS users upload. For this particular case, the image appears in different aspect ratios and it’s unlikely they will verify that the image they uploaded and cropped looks good on all screen sizes.

mmmavis: current prod code
image

mmmavis:
@Pomax I have images uploaded again!

Issue #770, opened 2020-01-21
Remove "ALL" currency from site


gideonthomas:
Braintree does not support Albanian Leks. Let's remove it as a currency option from the site.

TheoChevalier:
We will need to remove this line, too https://github.com/mozilla/donate-wagtail/blob/master/donate/payments/constants.py#L1074

Pomax:
If ALL is not supported, should we remove SQ as supported language, too, or should we instead default SQ to a different currency?

PR #827, opened 2020-02-13
remove ALL currency

Pomax:
closes https://github.com/mozilla/donate-wagtail/issues/770, if we're okay with removing SQ entirely.

If not, this needs a fix so that SQ does not default to `ALL` as currency.


TheoChevalier:
I’ve asked the localizer what he wants to do, it seems fair to be his decision

Pomax:
locale restored.

PR #848, opened 2020-02-21
Pass github token in headers and not as a query param

patjouk:
Using token in query params are not deprecated.

Related PRs/issues mozillafoundation/mofo-devops-private#297


donate/core/management/commands/review_app_admin.py
```diff
@@ -43,7 +43,9 @@ def handle(self, *args, **options):
                 token = settings.GITHUB_TOKEN
                 org = 'mozilla'
                 repo = 'donate-wagtail'
-                r = requests.get(f'https://api.github.com/repos/{org}/{repo}/pulls/{pr_number}&access_token={token}')
+                headers = {'Authorization': f'token {token}'}
+                r = requests.get(f'https://api.github.com/repos/{org}/{repo}/pulls/{pr_number}', headers=headers)
+                r.raise_for_status()
```
cadecairos: Did we not want to do anything with the exception raised here?

patjouk: It's raised, print something useful in logs, and fails the deployement of the review app.

cadecairos: Works for me :)

PR #855, opened 2020-02-28
Pontoon workflow follow-up

TheoChevalier:


tasks.py
```diff
@@ -89,10 +89,12 @@ def docker_makemigrations(ctx):
 @task
 def docker_makemessages(ctx):
     """Extract all template messages in .po files for localization"""
+    ctx.run("./translation-management.sh import")
```
patjouk: oopsie, forgot that in my previous PR. Sorry!

Issue #691, opened 2020-01-06
Add "no value selected" error state on landing page


alanmoo:
When landing on the homepage, attempting to choose a payment method without putting in a donation amount simply reloads the page with no visible error. Let's change that.

cadecairos:
In the old donate, we always had a value selected by default, we should do the same here.

Pomax:
As a not for whoever ends up doing this: this cannot be a `selected` attribute because of React, it needs to be a setState during componentDidMount, or preset state in the constructor.

Pomax:
Based on the preset-from-url work, this should be easy to do.

Pomax:
@WillatMozFdn I can put this back in, is there a specific amount (or nth button) that should preselected?

WillatMozFdn:
Let's go with the 2nd lowest value -- it's (generally) the closest to the average donation.  Thanks!

Pomax:
sweet, will do.

PR #902, opened 2020-03-25
ensure either amount= or second-lowest is preselected

Pomax:
Closes https://github.com/mozilla/donate-wagtail/issues/691 by making sure that if `amount` is specified, we use that IF that value exists in the list of values to choose from, otherwise (either it's not specified or it's not in the list) we preselect the second-lowest amount.


Pomax:
@cadecairos rereview requested, because it turns out I needed to update all the tests since this PR changes the "what default value do you get" behaviour. And disallows `1e100` etc =)

PR #954, opened 2020-05-06
Enable debugging in VSCode with ptvsd

cadecairos:
This PR enables VSCode to connect to a ptvsd server running in the docker container to enable interactive python debugging and breakpoints.

Instructions for using added to the Readme


.gitignore
```diff
@@ -97,10 +97,12 @@ ENV/
...
-/.vscode
 
 # PyCharm
 .idea/
 
 # Cypress visual regression testing
 cypress/videos
+
+# VSCode project config
+.vscode/settings.json
```
Pomax, line 108: It would be nicer if we could say "everything except for launch.json", e.g. https://stackoverflow.com/a/987162

cadecairos: made that change 👍

PR #960, opened 2020-05-12
Adding a language switcher

### LINKED ISSUE #961: Add a language switcher
Users should be able to manually change the site language via a language switcher in the footer, similar to the one on https://foundation.mozilla.org

Work happening in #960 

Asked multiple people for review in the PR, as it’s not a big or complicated PR, but touches different areas:

@mmmavis for SCSS/template code
@kristinashu for design
@Pomax /or @cadecairos for the Python changes

TheoChevalier:
Fixes #961 

This PR adds a language switcher in the footer, with CSS adapted from the foundation site

image

Test URLs:
- https://donate-wagta-language-s-ggd1y1.herokuapp.com/en-US/
- https://thunderbird-language-sw-yv8py8.herokuapp.com/en-US/

I’ve had to add additional metadata for locales not recognized by Django in base.py, and fixed some issues in the process


Pomax:
We're trying to move away from approving PRs without issues, so could I ask you to please file an engineering issue for this, with a story point estimation (1=trivial change, 2=maybe a few hours, 3=half a day, 5=a day, 8=2 days, anything above that is "this was an epic, not an issue), and put it in the current milestone (May 18). and then please update the comment to say "closes #...." so that we can keep properly tracking the work that has to be done, as well as the work done in past sprints.

TheoChevalier:
> We're trying to move away from approving PRs without issues, so could I ask you to please file an issue for this with the "engineering" label, with a story point estimation (1=trivial change, 2=maybe a few hours, 3=half a day, 5=a day, 8=2 days, anything above that is "this was an epic, not an issue), and put it in the current milestone (May 18). and then please update the comment to say "closes #...." so that we can keep properly tracking the work that has to be done, as well as the work done in past sprints.
> 
> (or of course if there _was_ an issue for this already, just say "closes #..." with that issue number)
> 
> For future reference: starting with writing the code solution is fine of course, but then once you're done and have it working, please file the issue it's going to close first, before you open the PR itself =D

There’s already a linked issue ;) https://github.com/mozilla/donate-wagtail/issues/961 I linked it in both GitHub and Zenhub, but maybe I missed something? I’ll look into the story point thing

Pomax:
hm, your initial comment still does not say "Closes #..." or "Fixes #...", which is the built in Github syntax for making sure than when this PR lands, the associated issue gets closed, which is pretty important =)

donate/templates/fragments/language_switcher.html
```diff
@@ -0,0 +1,23 @@
...
+      </svg>
+      {% trans "Language" %}
+    </label>
+    <select name="language" id="language-switcher" class="form-control" onchange="this.form.submit()">
+      {% get_current_language as LANGUAGE_CODE %}
+      {% get_available_languages as LANGUAGES %}
+      {% for CODE, NAME in LANGUAGES %}
+        {% get_language_info for CODE as lang %}
+        <option value="{{ CODE }}"{% if CODE == LANGUAGE_CODE %} selected{% endif %}>
+          {{ lang.name_local | capfirst }}
```
Pomax, line 18: just because I'm curious: is `capfirst` unicode aware?

source/sass/components/_footer-custom-select.scss
```diff
@@ -0,0 +1,36 @@
+.footer-links {
+  select {
+    &.form-control {
+      $background-size: 24px;
+      $padding-x: 12px;
+      $padding-right: $padding-x * 2 + $background-size;
+
+      -moz-appearance: none;
+      -webkit-appearance: none;
+      background-image: url("data:image/svg+xml,%3Csvg width='24' height='24' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M4 8l8.05 8.05L20.1 8' stroke='%23FFF' stroke-width='2' fill='none' fill-rule='evenodd' stroke-linecap='round' stroke-linejoin='round'/%3E%3C/svg%3E");
```
Pomax, line 10: This looks like it's the same SVG as the one added to the svg fragment file... if so, this is going to desync at some point: is there a way for us to make sure both files point to the same thing instead?

source/sass/components/_footer.scss
```diff
@@ -70,6 +70,46 @@
     background-color: $color--white;
   }
 
+  &__language-switcher {
+    width: 100%;
+    margin-bottom: 32px;
+    display: flex;
+    -webkit-box-orient: horizontal;
```
Pomax, line 77: As something that's got somewhat [wonky support](https://caniuse.com/#search=box-orient) we probably want to keep this simpler and just rely on flex?

TheoChevalier:
> hm, your initial comment still does not say "Closes #..." or "Fixes #..." (usually on its own line) which is the built in Github syntax for making sure than when this PR lands, the associated issue gets closed by Github. which is pretty important =)

It looks like manually connecting the two does that too (first time I’m using that feature)
image

source/sass/components/_footer-custom-select.scss
```diff
@@ -0,0 +1,36 @@
+.footer-links {
+  select {
+    &.form-control {
+      $background-size: 24px;
+      $padding-x: 12px;
+      $padding-right: $padding-x * 2 + $background-size;
+
+      -moz-appearance: none;
```
Pomax, line 8: Note that we're using [autoprefixer](https://github.com/mozilla/donate-wagtail/blob/master/package.json#L6) so there should not be any need to explicitly write vendor-prefixed rules in our SCSS

Pomax:
It's not a feature the rest of the engineering team uses, though, can you please still explicitly say which issue this closes in the first comment? It's nice if everyone follows the same practices so we can rely on seeing information where it's expected =(

donate/templates/fragments/language_switcher.html
TheoChevalier: The documentation doesn’t provide much detail, but from what I’ve seen, it seems so, yes — it’s smart enough to ignore cases where the concept of capital letter doesn’t exist, and does the right thing for cases like é → É

TheoChevalier:
Sure, I didn’t realize this was used to find the issue, not just get it closed automatically :)

source/sass/components/_footer-custom-select.scss
TheoChevalier: These are actually unrelated, one is a globe, this one is the down chevron (coming from https://github.com/mozilla/foundation.mozilla.org/blob/master/source/images/glyphs/down-chevron-dark-theme.svg)

source/sass/components/_footer-custom-select.scss
TheoChevalier: Correct 👍 replaced both with `appearance: none;` (the foundation website may warrant the same update)

source/sass/components/_footer.scss
TheoChevalier: Right, this was not needed, only kept `flex-direction: row;`

Pomax:
This looks like it has the same problem that @mavis was able to fix for the foundation site a while ago: the text coloring is such that in Chrome, the picker itself is white-on-white.

image

@mmmavis do you remember how you fixed that for foundation?

TheoChevalier:
Good catch, I didn’t test on Windows. Just found out the fix here https://github.com/mozilla/foundation.mozilla.org/pull/4247

TheoChevalier:
Just verified this is now fixed for Chrome on Windows

Pomax:
Nice!

One more thing: we probably want to change how we're sorting the languages here, because this list is doing some really odd sorting:

image

TheoChevalier:
> Nice!
> 
> One more thing: we probably want to change how we're sorting the languages here, because this list is doing some really odd sorting:
> 
> image
> 
> It's probably sorting on the english "concise" name, but that leads to F, H, M, B, I, T, L with unicode strewn amidst that based on the language pronunciations don't fit where they've been placed (日本語 is pronounced Nihongo, if we're showing the true label, a Japanese speaker would expect that in the "N" position, not the "J" position) - can we sort this unicodily, based on the true label (i.e. the one that users actually see in the list), rather than alphabetically based on the underlying English?

Right, the thing is Django doesn’t have pronunciation metadata/transliterated names (like `Nihongo` for Japanese), so we need to move to a manually curated list, and do some research for each locale using a non-Latin script. Not entirely convinced this is 100% reliable, but I took a stab at it.
I think the easiest way to implement this is to simply sort the `languages.py` file the way we want

> image
> 
> There's alignment issue. Could we make sure the links section is narrower with right edge aligned with the content below?

Nice catch! Added a fix for that

TheoChevalier:
Alright, I think this is ready for another look, but I’m not sure how to trigger the review apps though

Pomax:
the review apps auto-build, so usually it's just a matter of "having to wait 10 minutes tops"

Pomax:
hm, looks like for whatever reason the review app process did not kick in, so I told heroku to make new ones. Should be a few minutes

Pomax:
> Right, the thing is Django doesn’t have pronunciation metadata/transliterated names (like Nihongo for Japanese), so we need to move to a manually curated list

hmm, do we? Can we not sort on the presentation label? (sure, that puts 日本語 after all the ascii, but that's still fine)

TheoChevalier:
> hmm, do we? Can we not sort on the presentation label? (sure, that puts 日本語 after all the ascii and Latin-... characters, but that's probably fine)

The more I looked into this, the more I’m convinced we should go this way (alpha sort) rather than making up our own list, even if it made sense. I think this order is more common and less likely to throw people off.

TIL: there’s a word for sorting lists :) https://en.wikipedia.org/wiki/Collation
And in retrospect, for something that seems trivial at first, but [becomes complicated very quickly](http://www.unicode.org/reports/tr10/), I’m not surprised. And here I was, thinking I could use a `|dictsort:1` to magically sort everything in the template.

So, I reverted my changes to languages.py and sorted the list alphabetically, taking into account that:
- Some localized language names are capitalized, some others are not (this is not a bug). Here comes [case folding](https://www.w3.org/International/wiki/Case_folding). It goes a bit beyond just converting to lowercase, [the documentation](https://docs.python.org/3/library/stdtypes.html#str.casefold) explains this way better than I could.
- We also need to sort normalized strings, beyond case sensitivity (e.g. diacritics). Comparing strings following the NFD Unicode normalization form (Normalization Form Canonical Decomposition) will put `Česky` between `Català` and `Dansk`, otherwise it would be at the end of the languages using Latin script.

Finally, we need to sort/collate strings using [locale.strxfrm()](https://docs.python.org/3/library/locale.html?highlight=strxfrm#locale.strxfrm)

In code, this gives `return sorted(languages, key=lambda x: locale.strxfrm(unicodedata.normalize('NFD', x[1])).casefold())`

@Pomax could you have a final look at my last commit? (I promise I will stop making changes after that if it looks good 😅)

***
I also fixed a small visual issue happening when the label wrapped in multiple lines — the globe icon was shrunk to 16px (this was happening in Chinese)

Screenshot_2020-05-22 Donate today 捐款给 Mozilla

TheoChevalier:
Thanks to some great help from @patjouk , we found out that Heroku was using the `en_US.UTF-8` while Docker was using `C.UTF-8`. Not sure why en_US produces buggy results on Heroku, but forcing `C.UTF-8` everywhere works, so we did that.

`locale.setlocale(locale.LC_ALL, "C.UTF-8")`

On a different note, Mergify seem to have issues triggering Travis to merge

PR #995, opened 2020-05-27
Revert "Bump postcss-cli from 7.1.0 to 7.1.1 (#985)"

patjouk:
This reverts commit 38e5093781d541bbba457fb53ae1328f6ec51b11.

This is breaking on Heroku, but not locally or on Travis. Let's investigate why.


patjouk:
https://github.com/postcss/postcss-cli/releases/tag/7.0.0
version 7 dropped support for node 8, which we're still running. I'm closing this PR to open a new one that will pin this dependency to 6.1.3.

PR #1004, opened 2020-06-03
pass the fraud_donation_website custom var to braintree

### LINKED ISSUE #934: Pass custom variable to Kount to differentiate MoFo vs. T-bird transactions
Currently, we're unable to differentiate between MZLA and MoFo transactions in Kount, which creates difficulties for accounting (as Kount is a paid service) and also in donor care operations (MoFo staff are subsidizing MZLA with a LOT of time reviewing MZLA's suspect transactions in Kount).

Fortunately, Kount has the option for a "website" parameter that will help us sort and classify transactions.  This field needs to be passed in via a custom/hidden form field from our respective donation pages.  Three steps to the process:

1. Create the new custom field in Braintree (similar to how we have locale, campaign ID and project) ... instructions are here: https://articles.braintreepayments.com/control-panel/custom-fields .  It should be named "fraud_donation_website" and the value will be similar to/determined the same way as the "project" values (but shorter, see below).

2. Add this field as a hidden parameter on all our donate webpages.   However, values need to be SHORTER as there is an 8-character limit going into Kount.  "tbird" and "mofo" _should_ match the values Kount is now expecting.

3. The Kount part is already done; custom field of "donation_website" is already created, and waiting for values of "tbird" or "mofo" to be passed over.

cadecairos:
Closes #934

## Checklist

**Tests**
- [X] Is the code I'm adding covered by tests?

**Documentation:**
- [X] Is my code documented?

We must add the new custom field to the braintree accounts, otherwise things will fail. I'll update here when that's done.

### DIFF
--- a/donate/payments/tests/test_views.py
+++ b/donate/payments/tests/test_views.py
@@ -290,6 +290,7 @@ class CardPaymentViewTestCase(TestCase):
             'campaign_id': '',
             'project': 'mozillafoundation',
             'locale': 'en-US',
+            'fraud_donation_website': 'mofo'
         })
 
     def test_get_address_info(self):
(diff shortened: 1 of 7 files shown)


cadecairos:
`fraud_donation_website` has been added to the braintree sandbox and prod.

PR #1100, opened 2020-08-13
Removing support of 29 currencies

### LINKED ISSUE #1061: Remove 29 currencies from MoFo and MZLA-TBird donation forms
The admin & accounting expense of keeping up with these 29 currencies -- which, all told, comprise about 1% of our net donations -- is costing us more than we're raising.

So, we need to remove the following currency options from both Foundation and Thunderbird contribution forms.

Currencies to be removed:
AED
ARS
AZN
BAM
BDT
CLP
CNY
DZD
EGP
GEL
GTQ
HRK
IDR
ILS
KRW
LAK
LBP
MAD
MYR
PHP
QAR
RON
SAR
SGD
THB
TRY
UAH
YER
ZAR

This will affect some locales for the donation pages -- they'll need a new default currency.  Those choices (in most cases made by the affected community) are on the "Impacted Locales" worksheet here: https://docs.google.com/spreadsheets/d/1GV77O4U8P6bwXbrnALm26A7EVAoq1QAZB9goDk4HiFA/edit?usp=sharing .

TheoChevalier:
Closes #1061

# Done
- Removing all 29 currencies listed by Will. Dropdown is updated, as well as the currency list on Ways to give.
- Setting EUR as new default for `kab` and `hr` as per community feedback
- Removing `es-AR`, es-CL, `en-ZA`, `ms` as locales since they are not needed anymore.
- Enabled Wagtail Redirects module

# Remaining
~~- We may want to set-up redirects for the first 3 to avoid 404s from some hardcoded links here and there, even if the traffic shouldn’t be too high:~~
`/es-AR/*` → `/es/*`
`/es-CL/*` → `/es/*`
`/en-ZA/*` → `/en-US/*`
~~@cadecairos do you prefer doing that at the domain level, or in Django?~~

~~- Properly closing and deleting accounts in Braintree for both Mozilla and Thunderbird~~


cadecairos:
>     * We may want to set-up redirects for the first 3 to avoid 404s

I think we'll want to set up redirects using the CMS: https://docs.wagtail.io/en/v2.0/editor_manual/managing_redirects.html



>     * Properly closing and deleting accounts in Braintree for both Mozilla and Thunderbird

I don't think there's a cost associated with leaving them alone, so we could just leave them for now. If we must delete them I think production merchant accounts can only be created and destroyed by braintree support.

TheoChevalier:
> > ```
> > * We may want to set-up redirects for the first 3 to avoid 404s
> > ```
> 
> I think we'll want to set up redirects using the CMS: https://docs.wagtail.io/en/v2.0/editor_manual/managing_redirects.html
> 

Even better! I’ll test on staging

> > ```
> > * Properly closing and deleting accounts in Braintree for both Mozilla and Thunderbird
> > ```
> 
> I don't think there's a cost associated with leaving them alone, so we could just leave them for now. If we must delete them I think production merchant accounts can only be created and destroyed by braintree support.

Sounds good. And that matches my understanding from earlier convos with Will, as long as money is not flowing into these accounts, we’re good

TheoChevalier:
Didn’t realize the Redirect module needed to be enabled, so did that + set up redirects here https://donate-wagta-drop-curre-i43tw9.herokuapp.com/admin/redirects/

PR #1217, opened 2020-11-10
Update Wagtail to 2.11

kaedroho:


TheoChevalier (approved):
Kind of rubber-stamping this so we can properly test this work on staging

Issue #1187, opened 2020-10-13
German and French donors unable to donate via paypal


Pomax:
**Describe the bug**
German and French users (approximately 50 in recent ~8 weeks have written in about this!) are reporting failures when trying to donate via Paypal.

Poking around, we seem to notice that ... 
a) not affecting US donors
b) if you're already logged in to PayPal before you start the donation process, the problem seems to not happen
c) the problem occurs on "child pages" in Wagtail, but can't be reproduced on the main parent (default) page at https://donate.mozilla.org/

**To Reproduce**
Steps to reproduce the behavior (Firefox):
1a. Make sure you're logged out of PayPal.
1b. Go to https://donate.mozilla.org/de/jetzt-spenden/?subscribed=1&frequency=monthly&utm_source=newsletter-mofo&utm_medium=email&utm_campaign=SeptRECkicker&utm_term=5351122&presets=5,10,15,20&currency=eur 
2. Click on '5 EUR (pro monat)' button and then click on Paypal button
3. Log in to your paypal account in the pop-up window, and authorize the monthly payment.
4. Window closes
5. Failure message appears.

(screenshots removed -- please ask Will for access to the google doc.)

WillatMozFdn:
Developer console log of someone who experienced the problem.  [mozilla_donate_errorlog_github1187_2.txt](https://github.com/mozilla/donate-wagtail/files/5616921/mozilla_donate_errorlog_github1187_2.txt)

WillatMozFdn:
^^^ that guy is Hungarian.

WillatMozFdn:
Another clue from a donor ... I added some bolding on the key part in the middle:

"I clicked the link to donate as it showed at the bottom of the new tab 
page in firefox, and I landed there:
https://donate.mozilla.org/fr/dec2020/?utm_source=desktop-snippet&utm_medium=snippet&utm_campaign=giving-tuesday-dec2020&utm_term=24352&utm_content=REL&amount=5&frequency=single&currency=eur&presets=5%2C15%2C30%2C50
However.
Clicking "5€", "unique" (erm, or was it another word? "Single", perhaps? 
Not to have it automatically monthly, that was the thing), and clicking 
to pay with paypal, led to confusing results.
A paypal window popped up, in which I had to identify myself (sure), 
confirm the paid amount, etc, and click "Pay" (well, mine was in French, 
as I'm French, so it said "Payer")
And here's the bug: after I click to pay in the paypal pop-up window, 
[...]

PR #1256, opened 2020-12-07
Fix euro custom presets

Pomax:
Closes https://github.com/mozilla/donate-wagtail/issues/1187

initial STR based on master:
1. checkout master
2. `docker-compose up`
3. visit http://localhost:8000/de/?presets=5,10,15,20&currency=eur
4. inspect the "10 euro" form element, and observe that its associated `<input>` has a value `10,00` rather than the `10.00` it should be

fixed STR:
1. checkout this branch
2. `docker-compose up`
3. visit http://localhost:8000/de/?presets=5,10,15,20&currency=eur
4. inspect the "10 euro" form element, and observe that its associated `<input>` has a value `10.00` as it should be

The reason this happened is that we're using `Decimal` datatypes, which (unlike plain numbers) get localized during template rendering by Django. The immediate solution is to wrap these amounts in Django's `localize off` template block, with the follow-up being cleaning up any use of code for working with decimal values (since we don't _allow_ decimal amounts)


KalobTaulien (approved):
Nice!

PR #1267, opened 2020-12-14
Changed default values & sorting all presets

### LINKED ISSUE #1181: Standardize default amounts and ordering for primary donation currencies
**Describe the bug**
Right now our defaults for ordering suggested donation amounts are high-to-low, but this actually goes counter to research I just learned about from the NextAfter NIO Summit.  Also, our suggested amounts vary quite considerably in the primary seven currencies that constitute the bulk of our donations.  So we'd like to standardize these.

**Expected behavior**
Consistent suggested donation amounts, ordered low to high, in primary seven currencies.

**Additional context**
For the following seven currencies ...

USD
AUD
CAD
CHF
EUR
GBP
NZD

... we would like to order suggested donation amounts from low to high, and use the following values as suggested amounts:

One-time: 10, 20, 30, 60
Recurring: 5, 10, 15, 20

Thank you.

KalobTaulien:
Closes #1181 

## Criteria
- [x] 7 main currencies use `10, 20, 30, 60` for single payments. `5, 10, 15, 20` for monthly payments. 
- [x] All preset monthly/single values are sorted from lowest to highest 
- [x] Second lower value is highlighted by default still

### DIFF
--- a/donate/core/models.py
+++ b/donate/core/models.py
@@ -53,7 +53,12 @@ class DonationPage(Page):
 
     @cached_property
     def currencies(self):
-        return deepcopy(constants.CURRENCIES)
+        currencies = deepcopy(constants.CURRENCIES)
+        # Re-order currency `single` and `monthly` amounts
+        for currency in currencies:
+            currencies[currency]['presets']['single'].sort()
+            currencies[currency]['presets']['monthly'].sort()
+        return currencies
 
     def get_initial_currency(self, request):
         # Query argument takes first preference
@@ -90,9 +95,7 @@ class DonationPage(Page):
                 ]
                 if amount
             ]
-        except InvalidOperation:
-            return initial_currency_info
-        except ValueError:
+        except (InvalidOperation, ValueError):
             return initial_currency_info
 
         if not custom_presets:
--- a/donate/core/models.py
+++ b/donate/core/models.py
@@ -106,9 +106,7 @@ class DonationPage(Page):
 
         sorting = request.GET.get('sort', False)
 
-        if sorting == 'true':
-            custom_presets.sort()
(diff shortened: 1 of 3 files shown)


Pomax (changes requested):
if we're now sorting by default, https://github.com/mozilla/donate-wagtail/pull/1267/files#diff-f2e2f52c274fd525051162d4291d2af2b0105b61502a32b8762343f09285c071R109-R112 no longer needs to "sort if true" , it just needs to reverse sort if `reverse` is specified

KalobTaulien:
@Pomax Good catch, I hadnt thought of that. I touched that up if you want to give this a quick once-over.

donate/core/models.py
```diff
@@ -53,7 +53,12 @@ class DonationPage(Page):
 
     @cached_property
     def currencies(self):
-        return deepcopy(constants.CURRENCIES)
+        currencies = deepcopy(constants.CURRENCIES)
+        # Re-order currency `single` and `monthly` amounts
+        for currency in currencies:
+            currencies[currency]['presets']['single'].sort()
+            currencies[currency]['presets']['monthly'].sort()
+        return currencies
```
Pomax, line 61: This works, but it would probably be better if we sort `constants.CURRENCIES` when that gets loaded in, rather than resorting everytime `.currencies()` gets called?

Pomax, line 61: Filed https://github.com/mozilla/donate-wagtail/issues/1271 to follow up with potential improvements.

Pomax (approved):

donate/core/models.py
KalobTaulien: Thank you!

PR #1318, opened 2021-02-17
Upgrade Wagtail to 2.11.4

### LINKED ISSUE #1317: Upgrade to Wagtail 2.11.4
Now that 2.11.4 is out with the fix for alias pages being in draft, we should upgrade

TheoChevalier:
Closes #1317

Tested locally that it was fixing the draft mode issue for alias pages.

Note on PR: `pip-lock` is using a new format for requirements, so I performed a dry run in a first commit, then a second one with the actual Wagtail version bump to keep the diff readable.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -4,35 +4,78 @@
 #
 #    pip-compile dev-requirements.in
 #
-backcall==0.1.0           # via ipython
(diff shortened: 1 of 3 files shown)


Pomax (approved):
As far as I can tell locally, this works. Let's get it on staging and see what happens!

Issue #285, opened 2019-09-30
Field validation on custom amount field


alanmoo:
When choosing a custom amount but leaving the value empty, clicking on either payment button should show some sort of error state before attempting to process.

@beccaklam Do we have error states for this?

- [ ] Verify this exists on the second screen

beccaklam:
I don't think there's a comp for that specific error. But I've attached the general error state styling below. In this case we could use the red border around the custom amount form field and the text below it could say _'Please add an amount'_ or _'Amount cannot be 0'_ or _'Amount must be greater than $3'_'.

**Mobile:**
![Screen Shot 2019-09-30 at 2 17 16 PM](https://user-images.githubusercontent.com/20977376/65904746-11339d80-e38d-11e9-90c2-c558d2085aae.png)

**Desktop:**
![Screen Shot 2019-09-30 at 2 17 31 PM](https://user-images.githubusercontent.com/20977376/65904754-1690e800-e38d-11e9-9183-7edeb1050f00.png)

benhohner:
@kristinashu is this still relevant? Do we still want it to have this behaviour?

kristinashu:
Yes, it does seem like we should have an error message for these cases. @benhohner is it possible to be specific? I.e.

- If custom amount box is selected but no amount is entered the error would say 'Please add an amount' 
- If someone enters 0 then error would say 'Amount cannot be 0'
- If someone enters an amount less that $3 then it would say 'Amount must be greater than $3.00'

Or should we have a generic one for all these cases? Perhaps 'Please enter an amount greater than $3.00'

@WillatMozFdn do you have any feedback on this copy?

WillatMozFdn:
Hmmm.  I'm leaning more towards sticking with a generic error message here (and the current "There were some errors with your form. Please amend the fields highlighted below" is acceptable in that regard) ...  because it seems there are a LOT of cases we would need to plan for here.  a) Every currency has a different minimum value (that's coded in a python file somewhere) so we can't just say "less than $3", there would need to be 20 different variants of that, one for each currency we accept ... b) also now with Wagtail pages (this is new since this bug was first filed) we do not/can not accept entry of custom amounts that include decimals (that is, you cannot donate $3.50) so we'd need to program for that, and c) all these error messages would need to be localized and programmed into the system.

[...]

PR #1354, opened 2021-03-22
added new handler for other amount input which will display a error message which will display if the user enters a invalid value

### LINKED ISSUE #363: Create clear error state of 'Other Amount' <input>
### Issue: Donate home 'Other Amount' <input> has no error state for empty, invalid inputs.

error-id
**How to fix:**

- Please add clear error ui changes that align with payment form error states instead of page refreshing

====

More Info:

Comments: Sorry that gif is so slow, I have no clue why.
Resources: Related to #343
cc @philwakefield

danielfmiranda:
Closes #363 
Link to staging site: [here](https://donate-wagta-363-other--rbbudz.herokuapp.com)

## Checklist

**Documentation:**
- [x] Did I update the documentation?


danielfmiranda:
Hi @sabrinang,

Kalob and I have worked together to add a new error message whenever the user enters an invalid value. At the moment, it is just rendered as red text like so:

image

I was wondering if you can take a look and let us know if we should make any design changes to the error message. 

Thank you!

mmmavis:
(unassigned Sabrina as reviewer)

@danielfmiranda I'm tagging @kristinashu (design manager) here since it's not clear in the ticket which designer we should be getting design input from.

mmmavis:
GitHub tip:

For PR that is still in the working in progress state, it would be awesome if you could create a [draft pull request](https://github.blog/2019-02-14-introducing-draft-pull-requests/) and add `WIP` in the PR title and description so it's clear for other people to know that the PR is not yet ready for review. 🤓

danielfmiranda:
Thank you for the tip Mavis!!

kristinashu:
Sorry for the delay on this! The error state style should align with our existing styles. I.e. the outline color changes but doesn't "glow" like in your screen cap above. A nice bonus would be if you also add a warning icon like in the footer (this is better for color blind users who might not notice the color outline change).

Screen cap from prod https://foundation.mozilla.org/en/campaigns/facebook/
image

image

kristinashu:
I just found this ticket https://github.com/mozilla/donate-wagtail/issues/285 where the designer made a general comment about error states. I'd say consistency is the key.

Also noticed that the credit card payment detail page uses this outline style (no glow).
image

kristinashu (changes requested):
Please see design feedback above

benhohner:
@danielfmiranda We have an existing design system (chat with Mavis or someone from the Design team to learn more), meaning error states are probably implemented already somewhere in the codebase. Let's see if we can find those classes instead of implementing new ones to prevent code duplication! See https://foundation.mozilla.org/en/style-guide/

danielfmiranda:
Thank you for the heads up Ben!! Will take a look

source/js/components/validation.js
```diff
@@ -0,0 +1,42 @@
+// Handler for "other amount" inputs
```
danielfmiranda, line 1: Since both the one time/ monthly donation inputs have to do the same validation, I pulled that logic into its own function at the bottom of the file. 

I then added event listeners to both of the inputs, to call the validation function on blur.

donate/templates/fragments/donate_form.html
```diff
@@ -34,13 +34,17 @@
...
                     </div>
                     {% endfor %}
-
-                    <div class="donation-amount donation-amount--two-col donation-amount--other">
-                        <input type="radio" class="donation-amount__radio" name="amount" value="other" id="one-time-amount-other" autocomplete="off" data-other-amount-radio="">
-                        <label for="one-time-amount-other" class="donation-amount__label donation-amount-other__label" data-currency="">{% get_localized_currency_symbol initial_currency_info.code %}</label>
-                        <label for="one-time-amount-other-input" class="donation-amount__label--hidden">{% trans "Other amount" %}</label>
-                        <input type="number" class="donation-amount__input" id="one-time-amount-other-input" placeholder="{% trans 'Other amount' %}" min="{{ initial_currency_info.minAmount }}" max="10000000" data-other-amount="">
-                    </div>
+                    <div class="donation-amount--two-col">
```
danielfmiranda, line 37: Wrapped both the input and the error message in a new div to achieve the effect requested from Kristina:

image

source/sass/components/_donation-amount.scss
```diff
@@ -8,6 +8,13 @@
   &--other {
     display: flex;
     flex-direction: row;
+
+    &--container {
```
danielfmiranda, line 12: Added this so the input for monthly donations will remain the same size as the other donation choices on tablets and bigger screens

danielfmiranda:
Hi @kristinashu, I have made the requested changes, and the error state now matches the rest of the fields on the form

Hi @Pomax, I tagged you in this issue as I was wondering if I can get some feedback on the validation.js file, to double check it meets the teams standards

Thank you everyone!  :)

kristinashu:
Thank you Daniel!

The field now looks good on Safari. Chrome doesn't allow me to even enter non-numbers so I couldn't get it to show an error field. But Firefox seems to show both the outline and the glow, it should only show the outline. I'm on macOS v11.2.3.

image

kristinashu (changes requested):
Stilling showing the glow in Firefox.

danielfmiranda:
Thanks for the catch Kristina! Update is now on the staging site

kristinashu (approved):
Yay, the design looks good! Ready for code review.

Pomax:
One thing we may need to address, but that might need to be a new issue, is that the error messages moves the buttons for donating, so you can't just change the amount to the right amount, and click the "visa" or "paypal" buttons, because the moment you try to click them, they move:

screenshot 767
screenshot 766

source/js/components/validation.js
```diff
@@ -0,0 +1,42 @@
...
+  );
+  const monthlyOtherAmountInput = document.querySelector(
+    ".monthly-amount-other-input"
+  );
+  const oneTimeOtherAmountErrorMessage = document.querySelector(
+    ".error-message__one-time-other-amount"
+  );
+  const monthlyOtherAmountErrorMessage = document.querySelector(
+    ".error-message__monthly-other-amount"
+  );
```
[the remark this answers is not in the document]
danielfmiranda: Thanks for the catch Pomax! This and the other suggestions have been resolved in the latest commit, so I am marking these conversations as resolved

source/js/components/validation.js
```diff
@@ -0,0 +1,43 @@
...
+      const attemptedValue = monthlyOtherAmountInput.value;
+      inputValueCheck(attemptedValue, monthlyOtherAmountErrorMessage);
+    }
+  });
+}
+
+function inputValueCheck(valueToCheck, messageToUpdate) {
+  const number = parseFloat(valueToCheck);
+  // Checking if attempted value is a number and >= 2
+  if (!isNaN(number) && number && number >= 2) {
```
Pomax, line 34: note that we have _lots_ of different currencies, so while for US$ the minimum is 2, if someone picks a different currency, this logic needs to tap into that currency's minimum value instead.

Pomax (changes requested):

source/js/components/validation.js
danielfmiranda: Hi Pomax, thanks for the catch! I have updated this file to now validate that the attempted donation amount is between the input's minimum and maximum value. Anything else, and the error message "invalid amount entered" will appear, and a tooltip will also let the user know how to fix it (ex: "Value must be higher than x").

source/js/components/validation.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  });
+  monthlyOtherAmountInput.addEventListener("blur", (e) => {
+    if (document.querySelector(".monthly-donation-amount-radio:checked")) {
+      inputValueCheck(monthlyOtherAmountInput, monthlyOtherAmountErrorMessage);
+    }
+  });
+}
+
+function inputValueCheck(inputObjectToValidate, messageToUpdate) {
```
danielfmiranda, line 29: Updated this function to now receive the input object which is being updated, take the minimum/maximum value for that input, and validate that the attempted value is in between those parameters.

source/js/components/validation.js
```diff
@@ -0,0 +1,48 @@
...
+  const number = parseFloat(inputObjectToValidate.value);
+  const minimumAmount = parseFloat(inputObjectToValidate.min);
+  const maximumAmount = parseFloat(inputObjectToValidate.max);
+  // Checking if attempted value is a number and greater the inputs minimum amount / less than the maximum amount
```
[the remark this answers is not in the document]
danielfmiranda: Thanks Pomax! Removing the comment and resubmitting for review :)

Pomax (approved):

Pomax:
Nice work, in it goes!

PR #1356, opened 2021-03-25
Add Scout support

tomusher:
Adds Scout APM package and enables it as a Django app, pulling settings from environment variables.


KalobTaulien (approved):
LGTM. Before we get this merged, can we double check the evironment variables are set in prod?

PR #1377, opened 2021-04-02
added autocomplete=off to both  monthly and one-time donation radio buttons

### LINKED ISSUE #484: Possibility of accidentally donating one-time instead of monthly
STR: 

1. Choose a monthly donation amount
2. Click credit card button
3. Hit the browser back button (verified in Firefox 71.0b3)
4. Observe the monthly radio is selected, but one-time payment boxes are active. Choosing one of these and moving forward again starts a one-time gift flow.

Expected behavior:
Hitting the browser back button should (ideally) keep monthly selected, with monthly options. Alternatively: the correct radio should be selected for one-time gift.

@philwakefield This feels like something your team would have the context to address, as I know there were other issues around this.

danielfmiranda:
Closes #484 

## Checklist

**Changes in Models:**
- [x] [Are my changes backward-compatible

### DIFF
--- a/donate/templates/fragments/donate_form.html
+++ b/donate/templates/fragments/donate_form.html
@@ -8,13 +8,13 @@
         {% block donate_form_options %}
             <div class="tabs__option" role="presentation">
                 <label for="one-time" class="tabs__label tabs__item--selected js-tab-item" aria-controls="tab-panel-1" aria-selected="true" id="tab-1" data-label="Single">
-                    <input type="radio" class="tabs__radio-option" name="donation_frequency" id="one-time" value="one-time" {% if initial_frequency == 'single' %}checked{% endif %}>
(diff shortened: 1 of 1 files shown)


danielfmiranda:
Hi Everyone,  I was able to find a fix for the bug by setting the "autocomplete" flag on both of the inputs to false. 

According to the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete), firefox, which was the only browser experiencing the issue, has this set to True by default. 

I tried to look for a way for the page to remember whether the user selected monthly or one time donation, and I can see that the template is conditionally rendering the correct form by checking for a variable called "initial_frequency",  which seems to be created by a python function in the [backend](https://github.com/mozilla/donate-wagtail/blob/72e04d9926bc6f0281bfc3b92ef9177030544c66/donate/core/models.py). 

However, after looking at what this function returns for that value, no matter which frequency I select, it appears to always return "single".

So I believe that implementing that functionality would probably need its own ticket.

Thank you!

KalobTaulien (approved):

PR #1396, opened 2021-04-13
Send data to basket before sending to SQS

### LINKED ISSUE #1394: Change Donate Newsletter Subscriptions to use Basket HTTP API
Let's use the Python Basket Client if possible

https://pypi.org/project/basket-client/
https://basket-client.readthedocs.io/en/latest/

KalobTaulien:
Closes #1394 

Steps to test:
1. Make sure your braintree env vars are set. Ask me or Daniel for them if you need sandbox settings. 
2. Make a donation. 
3. On page 3 there's a "thank you" page. It requires a donation to be made before being able to view it. 
4. Add your email address, check the checkbox, and subscribe. We should see no errors. 

Env vars already set for Thunderbird (RA and Production) along with the main donate site (RA and Production) 

Review comments below.


dev-requirements.txt
```diff
@@ -54,28 +67,44 @@ ptyprocess==0.6.0
...
     # via
     #   -c requirements.txt
     #   requests
 wcwidth==0.1.9
     # via prompt-toolkit
-zipp==3.1.0
-    # via importlib-metadata
+zipp==3.4.1
+    # via
+    #   -c requirements.txt
```
KalobTaulien, line 106: All of these came from whatever was in master, and when I added `basket-client`

donate/payments/views.py
```diff
@@ -785,6 +787,8 @@ def form_valid(self, form, send_data_to_basket=True):
         data = form.cleaned_data.copy()
 
         if settings.POST_DONATE_NEWSLETTER_URL is not None:
+            # This will send the email address to Mailchimp for Thunderbird only.
+            # TODO: Check if this is even being used.
```
KalobTaulien, line 791: Not sure who to ask about this. We should double check with someone to make sure this is still being used for Thunderbird. I suspect it is, but it's good to know for sure.

donate/payments/views.py
```diff
@@ -797,7 +801,9 @@ def form_valid(self, form, send_data_to_basket=True):
                     exc_info=True
                 )
 
-        elif send_data_to_basket:
+        if send_data_to_basket:
```
KalobTaulien, line 804: Changed the `elif` to an `if`. This will affect Thunderbird. We are now sending subscription info from thunderbird-donate _and_ donate-wagtail to Basket.

donate/settings/base.py
```diff
@@ -37,6 +37,9 @@ class Base(object):
     # Basket Configuration
     BASKET_API_ROOT_URL = env('BASKET_API_ROOT_URL')
     BASKET_SQS_QUEUE_URL = env('BASKET_SQS_QUEUE_URL')
+    # Basket client configuration
+    BASKET_URL = env('BASKET_URL')
+    BASKET_API_KEY = env('BASKET_API_KEY')
```
KalobTaulien, line 42: Basket URL is set in the env.default file for the staging basket URL. 

The Basket API Key isn't needed for the standard HTTP request to Basket, but I put it in here for good measure in case we need it in the future.

donate/payments/views.py
```diff
@@ -797,7 +801,9 @@ def form_valid(self, form, send_data_to_basket=True):
...
+            basket.subscribe(data['email'], 'mozilla-foundation', lang=self.request.LANGUAGE_CODE)
+            # TODO: When going live we can remove the SQS commands below
             data['source_url'] = self.request.build_absolute_uri()
             data['lang'] = self.request.LANGUAGE_CODE
             queue.enqueue(send_newsletter_subscription_to_basket, data)
```
KalobTaulien, line 809: Note: I've kept SQS in here for the time being so this can be deployed without breaking the current site. We should remove this after the data migration work is complete.

danielfmiranda (approved):
Hi Kalob, 

Thanks for leaving some instructions in the comments as they were easy to follow when testing the new functionality. 

I took a look through the code and everything looks good to me, and I also followed the instructions and can confirm that there are no errors :) 

Regarding leaving the SQS functionality in there, I believe that was requested by @benhohner until we make the full switch over. 

Approving!

KalobTaulien:
@danielfmiranda I updated this to use an environment variable. Could you give the latest commit a review please?

.github/workflows/continuous-integration.yml
```diff
@@ -44,6 +44,7 @@ jobs:
       SLACK_WEBHOOK_PONTOON: ${{ secrets.SLACK_WEBHOOK_PONTOON }}
       SSL_REDIRECT: False
       USE_S3: False
+      DONATE_NEWSLETTER_SUBSCRIBE_METHOD: BASKET
```
KalobTaulien, line 47: Updated the env var name to be clearer for future tasks involving SQS. Thanks to @benhohner for this suggestion!

donate/payments/views.py
benhohner: This is definitely still being used, the env variable is set in thunderbird production and staging

donate/payments/views.py
benhohner: These options are all independent of each other for now.
I think the cleanest solution will be:
* Add an environment variable DONATE_NEWSLETTER_SUBSCRIBE_METHOD for donate staging and prod and thunderbird staging and prod
* Currently for donate:
    *  Staging:  DONATE_NEWSLETTER_SUBSCRIBE_METHOD = BASKET
    * Prod:  DONATE_NEWSLETTER_SUBSCRIBE_METHOD = SQS
* Currently for Thunderbird:
    * Staging:  DONATE_NEWSLETTER_SUBSCRIBE_METHOD = MAILCHIMP
    * Prod:  DONATE_NEWSLETTER_SUBSCRIBE_METHOD = MAILCHIMP

  Then for the logic
```python
if DONATE_NEWSLETTER_SUBSCRIBE_METHOD == "BASKET":
    # basket HTTP API code here
if DONATE_NEWSLETTER_SUBSCRIBE_METHOD == "SQS":
    # sqs code here
if DONATE_NEWSLETTER_SUBSCRIBE_METHOD == "MAILCHIMP":
    # mailchimp code here
```

donate/payments/views.py
benhohner: Let's separate it, see above. We don't want the same event sent twice as it'll end up in the same place

benhohner (changes requested):
Couple important changes so we don't send double events, see comments. Otherwise looks good!

donate/payments/views.py
```diff
@@ -798,15 +801,18 @@ def form_valid(self, form, send_data_to_basket=True):
...
-                    'eventLabel': 'Email',
-                }
-            ])
+            if settings.DONATE_NEWSLETTER_SUBSCRIBE_METHOD.lower() == 'basket':
+                basket.subscribe(data['email'], 'mozilla-foundation', lang=self.request.LANGUAGE_CODE)
+            else:
+                data['source_url'] = self.request.build_absolute_uri()
+                data['lang'] = self.request.LANGUAGE_CODE
+                queue.enqueue(send_newsletter_subscription_to_basket, data)
+                queue_ga_event(self.request, ['send', 'event', {
```
KalobTaulien, line 810: TODO: This needs to be moved outside the else statement. Always trigger a GA event.

donate/payments/views.py
```diff
@@ -798,15 +801,18 @@ def form_valid(self, form, send_data_to_basket=True):
...
-            data['source_url'] = self.request.build_absolute_uri()
-            data['lang'] = self.request.LANGUAGE_CODE
-            queue.enqueue(send_newsletter_subscription_to_basket, data)
-            queue_ga_event(self.request, ['send', 'event', {
-                    'eventCategory': 'Signup',
-                    'eventAction': 'Submitted the Form',
-                    'eventLabel': 'Email',
-                }
-            ])
+            if settings.DONATE_NEWSLETTER_SUBSCRIBE_METHOD.lower() == 'basket':
```
Pomax, line 804: if we're setting the value as `BASKET` in the env, and we default to `BASKET` in settings, let's compare to that, not to the lowercase string.

donate/payments/views.py
Pomax: If that still needs to happen, R- until that's done =)

Pomax (changes requested):

donate/payments/views.py
KalobTaulien: Done.

donate/payments/views.py
KalobTaulien: Done

KalobTaulien:
@benhohner @Pomax @danielfmiranda I updated the branch with the latest change requests. Let's re-review again.

danielfmiranda (approved):
Hi Kalob, just gave it another look after the updates you made with the feedback from others. 
Looks great to me, approving!

donate/payments/views.py
```diff
@@ -798,15 +804,22 @@ def form_valid(self, form, send_data_to_basket=True):
...
+            # During the switch phase, we're supporting both with an environment variable that
+            # determines where it should go.
+            if settings.DONATE_NEWSLETTER_SUBSCRIBE_METHOD == 'BASKET':
+                basket.subscribe(data['email'], 'mozilla-foundation', lang=self.request.LANGUAGE_CODE)
+            elif settings.DONATE_NEWSLETTER_SUBSCRIBE_METHOD == "SQS":
+                data['source_url'] = self.request.build_absolute_uri()
+                data['lang'] = self.request.LANGUAGE_CODE
+                queue.enqueue(send_newsletter_subscription_to_basket, data)
+
+        queue_ga_event(self.request, ['send', 'event', {
```
benhohner, line 817: I believe the original behaviour was to only send the event if send_data_to_basket is truthy. Based on the indent level of this function call, it looks like it's now outside the `elif send_data_to_basket` scope. Is this intended?

donate/settings/base.py
```diff
@@ -37,6 +37,10 @@ class Base(object):
...
```
benhohner, line 42: Now that we know more about the Basket client library, and that it doesn't require authentication, do you think we'll ever need to "use the lookup_user method with an email address"? (taken from basket client docs)

 I can't see any reason we'd need to do that in the near future, so I think it's safe to remove the BASKET_API_KEY.
```suggestion
```

donate/settings/environment.py
```diff
@@ -18,6 +18,8 @@
     AWS_STORAGE_BUCKET_NAME=(str, ''),
     BASKET_API_ROOT_URL=(str, ''),
     BASKET_SQS_QUEUE_URL=(str, ''),
+    BASKET_URL=(str, ''),
+    BASKET_API_KEY=(str, ''),
```
benhohner, line 22: See above comment
```suggestion
```

env.default
```diff
@@ -41,6 +41,11 @@ BRAINTREE_TOKENIZATION_KEY=
 # Basket
 BASKET_API_ROOT_URL=
 BASKET_SQS_QUEUE_URL=
+DONATE_NEWSLETTER_SUBSCRIBE_METHOD=BASKET
+
+# See the Basket client docs https://basket-client.readthedocs.io/en/latest/install.html
+BASKET_URL=https://basket-dev.allizom.org/
+BASKET_API_KEY=
```
benhohner, line 48:
```suggestion
```

benhohner (changes requested):
Just flagged a couple minor things. Looking good overall.

donate/payments/views.py
KalobTaulien: It was, but you raise a good point. It shouldn't be. I had overlooked that. Thanks for spotting this!

KalobTaulien:
@benhohner Solid change requests. I fixed those all up.

benhohner (approved):
R+ Thanks Kalob!

PR #1403, opened 2021-04-22
Removing new code in favour of old code

### LINKED ISSUE #1404: Use old basket code instead of new basket code.
We need to undo the code I wrote in favour of the existing code which uses rabbit mq for message persistence in case Basket isnt up so we dont lose the email address.

KalobTaulien:
Closes #1404 

Removing the new code in favour of the old code that uses rabbitmq to persist messages that may not make it to basket.


benhohner (approved):
R+. Sorry I didn't catch the duplicate code!

danielfmiranda (approved):
Hi Kalob, 

Just checked out this PR and everything looks good to me 👍 approving! 

PS: Thanks for taking the time to do this, I am sure this is going to make the code way easier to maintain in the future!

PR #1423, opened 2021-04-27
Updated how acoustic checked whether or not we are running tests

### LINKED ISSUE #1407: Research: Update email-sending during testing
At the moment, we have a env var called "is_testing" in continuous-integration.yml which is used to make the test suite skip sending a donation email while testing. 

While this works at the moment, we believe that there is a better way to implement this. Please research to see if there is a way we can keep the test suite from sending email receipts without the need of an env var

danielfmiranda:
Closes #1407 

Instead of adding a ENV variable called "IS_TESTING", we are now using the "TESTING" flag which is in the app's settings.
If the application is running its tests, emails will not be sent through acoustic, however if the app is deployed it will send emails through acoustic. 

## Checklist

**Changes in Models:**

### DIFF
--- a/donate/utility/acoustic/acoustic.py
+++ b/donate/utility/acoustic/acoustic.py
@@ -10,8 +10,6 @@ from silverpop.api import Silverpop, SilverpopResponseException
 
 logger = logging.getLogger(__name__)
 XML_HEADER = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
-# Checking whether or not we are running tests
-TESTING = settings.TESTING
 
 
 def process_response(resp):
@@ -117,7 +115,7 @@ class AcousticTransact(Silverpop):
 
     def send_mail(self, to, campaign_id, fields=None, bcc=None, save_to_db=False):
         # If we are testing, do not send emails
-        if not TESTING:
+        if not settings.TESTING:
             self._call_xt(transact_xml(to, campaign_id, fields, bcc, save_to_db))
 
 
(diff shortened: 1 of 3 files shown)


donate/utility/acoustic/acoustic.py
```diff
@@ -115,7 +117,7 @@ def _call_xt(self, xml):
 
     def send_mail(self, to, campaign_id, fields=None, bcc=None, save_to_db=False):
         # If we are testing, do not send emails
-        if not hasattr(settings, "IS_TESTING"):
+        if not TESTING:
```
danielfmiranda, line 120: instead of checking for the IS_TESTING variable, check the settings whether or not the TESTING flag is True

donate/utility/acoustic/acoustic.py
```diff
@@ -10,6 +10,8 @@
 
 logger = logging.getLogger(__name__)
 XML_HEADER = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
+# Checking whether or not we are running tests
+TESTING = settings.TESTING
```
benhohner, line 14:
```suggestion
```
I think we can call this directly from the conditional

donate/utility/acoustic/acoustic.py
benhohner, line 120:
```suggestion
        if not settings.TESTING:
```

Side note: I saw the previous version used `hasattr()`. For some extended reading, check out https://hynek.me/articles/hasattr/

benhohner (changes requested):

donate/utility/acoustic/acoustic.py
danielfmiranda: Hi Ben,

Thanks for the great feedback and reading material 👍, I have made the requested changes and re-requested review!

benhohner (approved):
R+

Pomax (approved):

PR #1437, opened 2021-05-04
updated CSP values in env.default

### LINKED ISSUE #1435: Update env.default to include CSP values
The https://github.com/mozilla/donate-wagtail/blob/master/env.default file currently does not come with preset CSP values similar ot the foundation's env.default, despite donate staging/prod having these values set.

We should ammend this file.

danielfmiranda:
Closes #1435 
Related PRs/issues #

## Checklist

**Changes in Models:**

### DIFF
--- a/env.default
+++ b/env.default
@@ -59,16 +59,15 @@ BASKET_SQS_QUEUE_URL=
 BASKET_API_ROOT_URL=https://basket-dev.allizom.org/
 
 # CSP config
-CSP_CHILD_SRC=" 'self' https://www.youtube.com https://www.youtube-nocookie.com "
 CSP_CONNECT_SRC=" * "
-CSP_DEFAULT_SRC=" 'none' "
-CSP_FONT_SRC=" 'self' https://fonts.gstatic.com https://fonts.googleapis.com https://code.cdn.mozilla.net "
-CSP_FRAME_ANCESTORS=" 'none' "
-CSP_FRAME_SRC=" 'self' https://www.youtube.com https://comments.mozillafoundation.org/ https://airtable.com https://docs.google.com/ https://platform.twitter.com https://public.zenkit.com https://calendar.google.com https://www.youtube-nocookie.com https://form.typeform.com "
+CSP_FONT_SRC=" 'self' fonts.googleapis.com fonts.gstatic.com"
+CSP_FORM_ACTION=" 'self' https://webto.salesforce.com/ mozillasalescloud--mkl.cs4.my.salesforce.com "
+CSP_FRAME_SRC=" 'self' assets.braintreegateway.com c.paypal.com *.paypal.com https://www.google.com/recaptcha/api2/ https://tst.kaptcha.com/ "
 CSP_IMG_SRC=" * data: "
-CSP_MEDIA_SRC=" 'self' https://s3.amazonaws.com/mofo-assets/foundation/video/ "
-CSP_SCRIPT_SRC=" 'self' 'unsafe-inline' https://www.google-analytics.com/analytics.js http://*.shpg.org/ https://comments.mozillafoundation.org/ https://airtable.com https://platform.twitter.com https://cdn.syndication.twimg.com https://embed.typeform.com "
-CSP_STYLE_SRC=" 'self' 'unsafe-inline' https://code.cdn.mozilla.net https://fonts.googleapis.com https://platform.twitter.com "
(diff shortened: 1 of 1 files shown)


env.default
```diff
@@ -58,6 +58,15 @@ BASKET_SQS_QUEUE_URL=
 # See the Basket client docs https://basket-client.readthedocs.io/en/latest/install.html
 BASKET_API_ROOT_URL=https://basket-dev.allizom.org/
 
+# CSP config
+CSP_CONNECT_SRC=" * "
+CSP_FONT_SRC=" 'self' fonts.googleapis.com fonts.gstatic.com"
+CSP_FORM_ACTION=" 'self' https://webto.salesforce.com/ mozillasalescloud--mkl.cs4.my.salesforce.com "
```
Pomax, line 64: Do we need both of these for the code to run correctly on localhost?

env.default
```diff
@@ -58,6 +58,15 @@ BASKET_SQS_QUEUE_URL=
...
+CSP_FRAME_SRC=" 'self' assets.braintreegateway.com c.paypal.com *.paypal.com https://www.google.com/recaptcha/api2/ https://tst.kaptcha.com/ "
```
Pomax, line 65: Do we need all of these for the code to run correctly on localhost? Also some of these don't have a protocol, let's make sure everything that is necessary to run locally also starts with https://

env.default
```diff
@@ -58,6 +58,15 @@ BASKET_SQS_QUEUE_URL=
...
+CSP_IMG_SRC=" * data: "
```
Pomax, line 66: Do we use data: urls for images on donate? If so, then this makes sense, but do we?

env.default
```diff
@@ -58,6 +58,15 @@ BASKET_SQS_QUEUE_URL=
...
+CSP_SCRIPT_SRC=" 'self' 'unsafe-inline' https://www.google-analytics.com/analytics.js  https://www.google-analytics.com/plugins/ua/ecommerce.js js.braintreegateway.com assets.braintreegateway.com www.paypalobjects.com c.paypal.com www.paypal.com https://www.google.com/recaptcha/api.js https://www.gstatic.com/recaptcha/releases/"
```
Pomax, line 67: Most of these look sensible, although we want some https:// prefixes for those bare domains.

env.default
```diff
@@ -58,6 +58,15 @@ BASKET_SQS_QUEUE_URL=
...
 BASKET_API_ROOT_URL=https://basket-dev.allizom.org/
 
+# CSP config
+CSP_CONNECT_SRC=" * "
+CSP_FONT_SRC=" 'self' fonts.googleapis.com fonts.gstatic.com"
+CSP_FORM_ACTION=" 'self' https://webto.salesforce.com/ mozillasalescloud--mkl.cs4.my.salesforce.com "
+CSP_FRAME_SRC=" 'self' assets.braintreegateway.com c.paypal.com *.paypal.com https://www.google.com/recaptcha/api2/ https://tst.kaptcha.com/ "
+CSP_IMG_SRC=" * data: "
+CSP_SCRIPT_SRC=" 'self' 'unsafe-inline' https://www.google-analytics.com/analytics.js  https://www.google-analytics.com/plugins/ua/ecommerce.js js.braintreegateway.com assets.braintreegateway.com www.paypalobjects.com c.paypal.com www.paypal.com https://www.google.com/recaptcha/api.js https://www.gstatic.com/recaptcha/releases/" 
+CSP_STYLE_SRC=" 'self' 'unsafe-inline' fonts.googleapis.com fonts.gstatic.com "
```
Pomax, line 68: Let's add some https:// here, too

Pomax (changes requested):

danielfmiranda:
Hi Pomax, 
As discussed in our call earlier, I have updated the env.default file to match the variable names in default.py, and also added  a comment to let readers know that they can find the default values in the .py file. Please let me know if there are any more requested changes or questions. 
Thanks!

env.default
```diff
@@ -58,6 +58,18 @@ BASKET_SQS_QUEUE_URL=
...
+# Default values for CSP items can be found in "donate/settings/defaults.py"
+CSP_DEFAULT_SRC=
+CSP_SCRIPT_SRC=
+CSP_IMG_SRC=
+CSP_FONT_SRC=
+CSP_STYLE_SRC=
+CSP_FRAME_SRC=
+CSP_CONNECT_SRC=
+CSP_BASE_URI=
+CSP_WORKER_SRC=
```
Pomax, line 71: we don't need these empty strings now. 
```suggestion
```

Pomax (changes requested):

env.default
danielfmiranda: Hi Pomax! Thanks for the heads up, I have removed them, and resubmitted for review 👍

Pomax (approved):

Pomax:
Sorry for the run-around! Turns out that the fix was actually just to make sure the "where to look" instructions are in a place where folks look first.

PR #1458, opened 2021-05-19
Backend postcode validation

### LINKED ISSUE #367: Zip code has no error for invalid input
### Issue: Zip code has no error for random, long input 
zip-code error

**How to fix:**

- Please add clear error logic that only accepts valid zip code format and flags anything else as an error
- Add error state styles as well (_although I assume this will kick in once logic is fixed_)

====

More Info:

Comments: n/a
Resources:
Related to: #343
Related to: Consider making postcode optional #415
cc @philwakefield

danielfmiranda:
Closes #367 

Review App can be found at: https://donate-wagta-backend-po-oqi62q.herokuapp.com/en-US/

**Steps to test country with post code:** 

1.) Visit [Donate Site Review App](https://donate-wagta-backend-po-oqi62q.herokuapp.com/en-US/)
2.) Select Canada/US/Other country that requires post code
3.) Fill out information for all fields under "Your Information" except ZIP/Postal Code
4.) Fill out card information: 
(Example Card Info, please feel free to use these):
Credit Card: 4111 1111 1111 1111
Exp: 11/23
CVV: 123

5.) Submit the Form
6.) The ZIP code input should now be displaying an error state (red border/red error message underneath)

**Steps to test country without post code:** 

1.) Visit [Donate Site Review App](https://donate-wagta-backend-po-oqi62q.herokuapp.com/en-US/)
2.) Select Tonga/Other country that does not require post code
3.) The ZIP field should now dissapear, just leaving the city field.
3.) Fill out information for all fields under "Your Information" 
4.) Fill out card information: 
(Example Card Info, please feel free to use these):
Credit Card: 4111 1111 1111 1111
Exp: 11/23
CVV: 123

5.) Submit the Form
6.) You should see the thank you page

And thats it! Thank you for taking the time to test
 

## Checklist

**Changes in Models:**


donate/templates/payment/card_master.html
```diff
@@ -74,8 +74,12 @@ <h1 class="heading heading--primary heading--bottom-margin">{% trans "Your Detai
                             {% render_form_field form.last_name %}
                             {% render_form_field form.address_line_1 %}
                             <div class="form__group form__group--city-post">
-                                {% render_form_field form.post_code %}
-                                {% render_form_field form.city %}
+                                <div class="post-code-input-container">
```
danielfmiranda, line 77: Wrapped these fields in respective divs so I can add the "hidden"/"full-width" class when neccesary

source/sass/components/_form.scss
```diff
@@ -39,12 +39,27 @@
...
+            border-left: 0;
+          }
+        }
+      }
+
+      &--full-width {
+        grid-template-columns: 1fr;
+        .city-input-container {
+          .form-item {
+            input {
```
danielfmiranda, line 54: Since in the original 2 column format the border-left is set to 0, we are setting the " input field border to match the rest of the fields when in the full width format.

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
...
+
+// The container that houses both the post code and city inputs
+const formAndCityContainer = document.querySelector(".form__group--city-post");
+
+// The select object that allows users to select their country
+const countrySelector = document.getElementById("id_country");
+
+// Calling the function once to check in case the user refreshes the page,
+// we want the input to remain one column if needed
+if (countrySelector) {
```
danielfmiranda, line 19: The if statements on line 19 and 24 are verifying that the country select input exists on the page, if not, no need to run the function or add event handler yet

source/js/components/validation.js
```diff
@@ -14,16 +14,26 @@ const monthlyOtherAmountErrorMessage = document.querySelector(
...
-    if (document.querySelector(".one-time-donation-amount-radio:checked")) {
-      inputValueCheck(oneTimeOtherAmountInput, oneTimeOtherAmountErrorMessage);
-    }
-  });
-  monthlyOtherAmountInput.addEventListener("blur", (e) => {
-    if (document.querySelector(".monthly-donation-amount-radio:checked")) {
-      inputValueCheck(monthlyOtherAmountInput, monthlyOtherAmountErrorMessage);
-    }
-  });
+  if (oneTimeOtherAmountInput) {
```
danielfmiranda, line 17: added if statements here as well, as we found that this was also experiencing the same issue where it would throw an "undefined" error if we were not on the page that had these input fields

kristinashu (approved):
Thank you for the review app link and very clear steps on how to test. All worked as expected!

donate/payments/constants.py
```diff
@@ -528,3 +528,1247 @@
   'XOF',
   'XPF'
 ]
+
+
+# List of countries and their respective post code formats if applicable.
+# Used in post-code-validation.js to check whether or not we display a post code form item to the user.
+# Source of data can be found here: https://gist.github.com/lkopocinski/bd4494588458f5a8cc8ffbd12a4deefd
+# NOTE: This list is also found in donate/constants.py, any updates to this list should be made there/vice-versa.
+COUNTRY_POST_CODES = [
```
Pomax, line 537: Both Python and JavaScript can load in .json data, so if we need this list to be synced between the two, let's just use the JSON from the gist, and have both load that in, instead of having two files that can go out of sync.

(but remember that json parsing usually needs a try/catch or try/except)

donate/payments/forms.py
```diff
@@ -48,6 +48,31 @@ def clean(self):
...
+        country = cleaned_data.get('country', '')
+        # Getting countries post code information from list in constants.py
+        country_object_to_check = next(country_obj for country_obj in
+                                       constants.COUNTRY_POST_CODES
+                                       if country_obj["abbrev"] == country)
+        # If country uses a post code and the user left it blank, raise a validation error.
+        if 'postal' in country_object_to_check and postal_code == "":
+            raise forms.ValidationError({
+                'post_code': _('This field is required.')
+            })
```
Pomax, line 73: Remember the general rule for comments: if the comment says the same thing that the code does, it doesn't need to be a comment. In this case you may want a comment to explain why the coded uses `next(...)` but the rest of the code is clear enough on its own.

donate/templates/payment/card_master.html
Pomax: This should not be necessary: the `render_form_field` tag lets you specify additional custom classes, if needed (see https://github.com/Tivix/django-common/blob/407d208121011a8425139e541629554114d96c18/django_common/templatetags/custom_tags.py#L50-L73)

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
+import { countriesAndPostCodes } from "./post-codes-list.js";
+
+// List of countries to use for reference
+const listOfCountries = countriesAndPostCodes;
```
Pomax, line 4: We're not modifying `countriesAndPostCodes`, so we can probably do without aliassing.

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
...
+// The container that houses both the post code and city inputs
+const formAndCityContainer = document.querySelector(".form__group--city-post");
+
+// The select object that allows users to select their country
+const countrySelector = document.getElementById("id_country");
+
+// Calling the function once to check in case the user refreshes the page,
+// we want the input to remain one column if needed
+if (countrySelector) {
+  checkForCountrysPostCode();
```
Pomax, line 20: just for ease of reason, let's make this `checkForCountryPostCode()`, since `countrys` looks like an improper plural and may cause typo bugs in the future.

(and let's do that for the other instances too)

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
...
+
+  // Finding the country object in the array.
+  const countryObject = listOfCountries.find(
+    (country) => country.name === selectedCountrysName
+  );
+
+  // If the country exists in the array, check whether or not it has a post code.
+  if (countryObject !== undefined) {
+    // If the country does NOT have a post code, remove the zip code field.
+    if (!countryObject.hasOwnProperty("postal")) {
```
Pomax, line 44: unless you're trying to bypass prototype inheritance, there is no need to use `hasOwnProperty`. You can simply use `if (!countryObject.postal) { ... } else { ... }`, but it's better to switch those cases around: "if X, then Y, else Z" rather than "if not X, then Y, else Z", so:
```
if (countryObject) {
  if (countryObject.postal) {
    ...
  } else {
    ...
  }
}
```

Pomax (changes requested):

donate/templates/payment/card_master.html
Pomax: My bad - this is our own tag in `donate/core/templatetags/form_tags.py`, so we can just update that so that it can do additional classes.

danielfmiranda:
Hi @Pomax thanks for the detailed feedback and code review, and also for helping me find the right way to add custom classes to form fields! I have made the changes you requested, and re submitted the code for review. Please let me know if you have any questions. 

Thank you!

donate/templates/forms/form_field.html
```diff
@@ -7,6 +7,7 @@
 {% if field_type %}form-item--{{ field_type }}{% endif %}
 {% if widget_type %}form-item--{{ widget_type }}{% endif %}
 {% if field.errors %}form-item--errors{% endif %}
+{% if css_classes %} {{css_classes}} {% endif %}
```
Pomax, line 10: Just making sure we match the syntax/formatting that's already there:
```suggestion
{% if css_classes %}{{ css_classes }}{% endif %}
```

donate/templates/payment/card_master.html
```diff
@@ -74,7 +74,7 @@ <h1 class="heading heading--primary heading--bottom-margin">{% trans "Your Detai
                             {% render_form_field form.last_name %}
                             {% render_form_field form.address_line_1 %}
                             <div class="form__group form__group--city-post">
-                                {% render_form_field form.post_code %}
+                                {% render_form_field form.post_code css_classes="post-code-input"%}
```
Pomax, line 77:
```suggestion
                                {% render_form_field form.post_code css_classes="post-code-input" %}
```

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
+// Importing JSON list of countries and post-code info.
+// This info is also shared with donate/forms.py to check on the backend.
+try {
+  var countriesAndPostCodes = require("./post-codes-list");
+} catch (err) {
+  console.error(
+    "Could not import post-code data. Zip code is required by default."
+  );
+}
```
Pomax, line 9: This is ES module code, no `require` allowed. However, the `import` statement will work for json files:
```suggestion
import countriesAndPostCodes from "./post-codes-list.json";
```

source/js/components/post-code-validation.js
Pomax: If so, you'll want to move Lines 11 and 14 to inside the `if` block, because there's no reason to query for them if countrySelector is nothing.

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
...
+const formAndCityContainer = document.querySelector(".form__group--city-post");
+
+// The select object that allows users to select their country.
+const countrySelector = document.getElementById("id_country");
+
+// If you are on the form page, check if users country uses post code.
+if (countrySelector) {
+  checkForCountryPostCode();
+}
+// Event listener for checking post-code when user selects new country.
```
Pomax, line 23:
```suggestion
}

// Event listener for checking post-code when user selects new country.
```

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
...
+function countryPostCodeValidation() {
+  if (countrySelector && countriesAndPostCodes) {
+    countrySelector.addEventListener("change", (e) => {
```
Pomax, line 26: If we're not using the event, no need to explicitly capture it as call argument.
```suggestion
    countrySelector.addEventListener("change", () => {
```

source/js/components/post-code-validation.js
```diff
@@ -0,0 +1,58 @@
...
+  const countryObject = countriesAndPostCodes.find(
+    (country) => country.name === selectedCountrysName
+  );
+
+  if (countryObject !== undefined) {
+    if (countryObject.postal) {
+      // Display post code field.
+      formAndCityContainer.classList.remove(
+        "form__group--city-post--full-width"
+      );
```
Pomax, line 47: what does this line do, and is there a different way we can achieve the same thing? The fact that it hardcodes a CSS class including the words "full width" feels like it's one template/scss change way from being a bug we need to hunt down =)

Pomax (changes requested):

source/js/components/post-code-validation.js
danielfmiranda: Awesome catch Pomax!

source/js/components/post-code-validation.js
danielfmiranda: Hi Pomax! I have removed any of the hardcoding of css classes, and instead  change the css value of the grid in the for the container from "0.3fr 0.7fr" to "auto" to allow the city input to take the full width of the form.

Pomax (changes requested):
The code works quite well now, although I noticed two things while using it:
1. picking a country without a postal code does not clear the postal code input element. To prevent an unrelated postal code from being sent, let's make sure we also clear the postal code field when we hide it, and
2. it looks like there are some "non-participatory" countries: US, then Aland Islands, then US, the postal code is visible all three times. However, if I pick Tonga, Aland Islands, Tonga, the postcal code is _not_ visible all three times, so something funny is going on for the Aland Islands. 

It feels like we want to make sure (1) is addressed in this PR, but for (2) we can probably file a follow-up issue instead, as it seems such an extreme edge case that we can probably defer working on it for now.

danielfmiranda:
Hi Pomax! Thanks for the suggestion about clearing the inputs value if the user selects a country with no post code, as I can see that being very helpful to ensure that we do not submit random values into braintree in the future.

I have made that change, as well as updated the classname of the container that houses the zip-code input, in an attempt to make the code a little easier to see that we are hiding the container and everything in it, but removing the value from the actual input. 😄

Pomax (approved):
Everything looks good now. I've filed https://github.com/mozilla/donate-wagtail/issues/1477 for the followup investigation noted as point (2) in my comment, above.

PR #1460, opened 2021-05-20
Bumped Django and AB Testing

### LINKED ISSUE #1457: 🔥 Donate deploys are hitting their memory limit 🔥
The donate staging and production deployments are consistently riding above the 512MB memory limit, and we should figure out why, because it's adversely affecting service, which adversely affects our bottom line =)

looking at Scout, there are some calls that apparently lead to 80MB in memory allocations, let's figure out what those are?

KalobTaulien:
Partially Closes #1457 

**Note**: There was a missing migration. I ran migrations and plopped it in this branch to pass CI. It was a small migration. 
**Note:** We also can't upgrade to Wagtail 2.12 yet because we have packages that specify wagtail<2.12. Those will need version bumps because we can upgrade to Wagtail 2.12 and 2.13. 

## Testing steps 
1. Checkout this branch 
2. run `inv new-env`. No problems should arise. 
3. Poke around on the localhost website to ensure things are working smoothly.


danielfmiranda (approved):
Hi Kalob, 

Thanks for leaving some instructions on how to test this change. 
I pulled the branch and then ran inv new-env, and can confirm that everything was working the way they should be. 

Approving!

Issue #1472, opened 2021-05-27
Donate is logging BT calls


Pomax:
These things are huge, 100,000 character log entries that should absolutely not be anywhere near a persistent/long-term log. Let's find out where it's generating these, and kill them off.

Pomax:
Log entry:
```
2021-05-27T19:55:41.127470+00:00 app[worker.1]: 19:55:41: default: donate.payments.tasks.process_webhook({'bt_signature': ACBDEF, 'bt_payload': ZYXWVU}) (4e1aa963-6ade-41ac-92a1-3b4e67c4e679)
```
Code dive:
- `process_webhook` calls `notification = gateway.webhook_notification.parse(form_data['bt_signature'], form_data['bt_payload'])`
- `process_webhook` called by `BraintreeWebhookView`:
```
@method_decorator(csrf_exempt, name='dispatch')
class BraintreeWebhookView(FormView):
    form_class = WebhookForm
    http_method_names = ['post']

    def form_valid(self, form):
        queue.enqueue(process_webhook, form.cleaned_data)
        return HttpResponse()

    def form_invalid(self, form):
        return HttpResponseBadRequest()
```

- `queue.enqueue(process_webhook, form.cleaned_data)` is a redis queue defined in tasks.py:

```
queue = django_rq.get_queue('default')
```

- django_rq uses its own custom logging, documented in https://github.com/rq/django-rq#configuring-logging
- rq logging that we're using in defaults.py:

```
LOGGING = {
    'version': 1,
    'filters': {
        'require_debug_true': {
            '()': 'django.utils.log.RequireDebugTrue',
        }
    },
    'formatters': {
        'verbose': {
            'format': '%(asctime)s [%(levelname)s] %(message)s'
        },
        'rq_console': {
            'format': '%(asctime)s %(message)s',
            'datefmt': '%H:%M:%S'
        }
    },
    'handlers': {
        'debug': {
            'level': 'DEBUG',
            'class': 'logging.StreamHandler',
            'filters': ['require_debug_true'],
            'formatter': 'verbose'
        },
        'error': {
            'level': 'ERROR',
            'class': 'logging.StreamHandler'
        },
        'debug-error': {
            'level': 'ERROR',
            'filters': ['require_debug_true'],
            'class': 'logging.StreamHandler'
        },
        'info': {
            'level': 'INFO',
            'class': 'logging.StreamHandler',
            'formatter': 'verbose'
        },
        'rq_console': {
[...]

PR #1473, opened 2021-05-27
Update defaults.py to make sure rq_console is debug-only

Pomax:
Related PRs/issues #1472


Pomax:
@KalobTaulien I'm not actually sure how we can test this, short of "landing it and seeing if that does the trick"...

KalobTaulien (approved):

PR #1483, opened 2021-06-03
Message id key error fix

### LINKED ISSUE #1482: KeyError when getting message_id in receipt sending
**Describe the bug**
There seems to be an issue selecting a message_id for donation receipts. below is a screenshot from Sentry:

image

danielfmiranda:
Closes #1482 

We are now setting a default email language id in base.py, and if the app is not able to find a language it will immediately prevent the app from running. 

I also updated the function to use single quotes instead of double quotes. 

## Checklist

**Changes in Models:**

### DIFF
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -1,3 +1,4 @@
+from donate.settings import languages
 import logging
 import time
 from decimal import Decimal
@@ -17,7 +18,6 @@ from .utils import (
     get_plan_id,
     get_merchant_account_id_for_card
 )
-from ..settings.languages import LANGUAGE_IDS
 
 logger = logging.getLogger(__name__)
 
@@ -72,6 +72,8 @@ DONATION_RECEIPT_FIELDS_MAP = {
     "locale": "donation_locale",
     "service": "payment_source",
 }
+
+LANGUAGE_IDS = languages.LANGUAGE_IDS
 
 
 # Acoustic receipt sending
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -17,6 +17,7 @@ from .utils import (
     get_plan_id,
     get_merchant_account_id_for_card
 )
+from ..settings.languages import LANGUAGE_IDS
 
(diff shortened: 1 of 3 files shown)


donate/payments/tasks.py
```diff
@@ -73,48 +73,39 @@ def mofo_donation_receipt_number_format(amount):
     "service": "payment_source",
 }
 
-# These are compared with the donating users locale, to tell the Acoustic API what language to send the email
```
danielfmiranda, line 76: moved this array to languages.py, so we can use it in both this file and in base.py where we are setting defaults

donate/payments/tasks.py
```diff
@@ -17,6 +17,7 @@
     get_plan_id,
     get_merchant_account_id_for_card
 )
+from ..settings.languages import LANGUAGE_IDS
```
Pomax, line 20: you should never import settings as a file, because there is no guarantee you're loading the right one. Instead, ask django its current config:

```suggestion
from dangjo.conf import settings
```

and then you can safely extract anything from there:

```
LANGUAGE_IDS = settings.languages.LANGUAGE_IDS
```

Pomax (changes requested):
this just needs the one fix.

donate/payments/tasks.py
danielfmiranda: Hi @Pomax thank you for the catch and for the suggestion. 

One thing I did note was that referring to language id's through:

`settings.languages.LANGUAGE_IDS`
`settings.LANGUAGE_IDS`
`languages.LANGUAGE_IDS`

using the suggested method did not work and returned an error stating that "settings object has no attribute 'languages'".

Would defining the language_ids in the base.py file as the current PR is doing, and then referring to them as:
`LANGUAGE_IDS = settings.LANGUAGE_IDS` work, Or is there a downside in having to define language_ids in base.py?

Pomax (approved):

Pomax:
looking good, let's get it landed =)

PR #1485, opened 2021-06-04
Don't require valid SSL on redis connections

tomusher:


tomusher:
@KalobTaulien @pomax Could we get this one reviewed and merged soon? It is required for the Redis 6.0 upgrade that Heroku is going to make mandatory soon.

KalobTaulien (approved):

KalobTaulien:
Done @tomusher 

If you're able to you should be able to set me or Pomax as reviewers on the right side of the page. It'll send us a notification so we dont miss important things like this :) 

Let's deploy this next week sometime (not today)

PR #1498, opened 2021-06-14
Use the mailchimp-marketing API for thunderbird

### LINKED ISSUE #1471: [Thunderbird] After donating, when opting in for email updates, the server timesout and signups arent submitted
Original ticket edited: 

**Describe the bug**
When submitted a donation to Thunderbird, and then signing up for the newsletter, the page timesout after 30s and leaves the user with a "maintenance" page. 

**To Reproduce**
1. Donate to Thunderbird
2. Opt in (check the box) for email updates
3. Submit form
4. Should subscribe to Mailchimp (but takes a while to process) but ends up going to a maintenance page that says "Thank you for your interest" 

#### First
- [x] Check if this bug still exists. 
#### Then 
- [ ] ~The maintenance page _really_ should not say "Thank you for your interest" as that can be quite misleading.~
- [ ] We need to figure out why the request timed out and fix it. 
  - [ ] (Could be Mailchimp taking forever or could be our backend taking a long time) 
- [x] @benhohner to test this out in production

KalobTaulien:
Closes #1471 

## Steps for testing:
1. Checkout this branch
2. `inv new-env` (needed for the Mailchimp API package) 
3. Set your environment to be ThunderbirdDevelopment 
4. Add the `THUNDERBIRD_MC_API_KEY`, `THUNDERBIRD_MC_SERVER` and `THUNDERBIRD_MC_LIST_ID` to your local development settings 
5. Make a donation. It should no longer timeout (it should be nearly instant now) 
6. Make the same donation with the same email address, there should be no error (the mailchimp-marketing client will through an ApiClientError if the email address is already subscribed

If you want to test that it's actually works on local, you can add print statements between lines 788-814 in `donate/payments/views.py`


danielfmiranda (approved):
Hi Kalob, Thanks for the easy to follow instructions, I tested this on my local machine using the print statements you have mentioned and can confirm that it works and also throws an error when you sign up with the same email twice. Approving!

PR #1506, opened 2021-06-24
updated "service" to "payment method" in process_donation_receipt

### LINKED ISSUE #1505: Paypal receipts are missing donation value
**Describe the bug**
If a user makes a donation via paypal, the donation receipt will not have a donation value on the "via:" line. 

**To Reproduce**
1. Go to donation app homepage on local machine
2. Make a donation via paypal, either via sandbox paypal account or credit card (reach out to Will Easton for a refund). 
3. Note that the donation value is missing. 

**Expected behavior**
The donation amount should be displayed.

danielfmiranda:
Closes #1505 

It turns out the email function was looking for something with the key of "payment_method" but we were sending it "service". After making this change in the email function, the emails now differentiate between paypal payments and credit card payments! 

## Checklist

**Changes in Models:**

### DIFF
--- a/donate/payments/tasks.py
+++ b/donate/payments/tasks.py
@@ -80,8 +80,7 @@ def process_donation_receipt(donation_data):
     # Creating new object by looping through mandatory receipt fields from const dictionary,
     # and updating them to equal the data being received
     message_data = {k: v for k, v in donation_data.items() if k in DONATION_RECEIPT_FIELDS}
-    email = "daniel@mozillafoundation.org"
-    # email = message_data.pop('email')
+    email = message_data.pop('email')
     # If the donation data did not recieve a payment time, use the current time.
     created = message_data.get('created', int(time.time()))
     # The next 3 lines are formatting the date and time for the email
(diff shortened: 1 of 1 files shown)


donate/payments/tasks.py
```diff
@@ -60,16 +60,16 @@ def mofo_donation_receipt_number_format(amount):
     "card_type",
     "last_4",
     "locale",
-    "service",
```
danielfmiranda, line 63: The emailing receipt function was expecting to receive the payment method in a variable called service, however, it was sending it titled as "payment_method" instead. Have tested on local machine with Credit Card and Paypal and can confirm it is updating correctly now!

danielfmiranda:
Hi Everyone, currently working with Stephanie Mcvicker to get access to the test-paypal email account so we can verify that email sending is working correctly.

danielfmiranda:
Hi Everyone, after working on finding a way to access the emails for the test paypal account, it appears that there is currently no way. 

However, here is an example email of me using the test-paypal account but having the application send the email to me instead. 

Email From Donation with paypal


Since this fix was just a matter of changing a key from "service" to "payment_method" Ben and I think it is safe to assume that the fix is working, and that we can deploy and then verify it works in production.

KalobTaulien (approved):

PR #1507, opened 2021-06-25
updated braintree subscriptions to start immediately

### LINKED ISSUE #1500: Delay in receipt sending past a certain hour
**Describe the bug**
According to some of the people testing the new receipt sending, there is a delay in receiving the thank you receipt especially if you make a donation in the evening. 

**Expected behavior**
The thank you receipt should be sent shortly after the user has made the donation. 

Possible Causes: 
Our timezone on the server and braintrees timezone?

IIRC this issue has been resolved, however we are creating this ticket to verify.

danielfmiranda:
Closes #1500 

**Testing:**
Testing may be a little tricky on this PR since the webhook is tied to our staging app. 
If my suggested changes make sense however, I can submit a recurring donation at around 4:30 PM PST as that is when Jackie and Will Easton are claiming there is a delay on receipt sending, and can submit/confirm if the delay persists, on this ticket.

## Checklist

**Changes in Models:**

### DIFF
--- a/donate/payments/tests/test_views.py
+++ b/donate/payments/tests/test_views.py
@@ -1,4 +1,3 @@
-from datetime import date
 from decimal import Decimal
 from unittest import mock
 
@@ -450,7 +449,9 @@ class MonthlyCardPaymentViewTestCase(CardPaymentViewTestCase):
             'merchant_account_id': 'usd-ac',
             'payment_method_token': 'payment-method-1',
             'price': 50,
-            'first_billing_date': date(2019, 7, 26),
+            'options': {
+                "start_immediately": True
+            }
         })
 
         self.assertEqual(self.request.session['landing_url'], self.form_data['landing_url'])
@@ -597,7 +598,9 @@ class PaypalPaymentViewTestCase(TestCase):
             'merchant_account_id': 'usd-ac',
             'payment_method_token': 'payment-method-1',
             'price': Decimal(10),
-            'first_billing_date': date(2019, 7, 26),
+            'options': {
+                "start_immediately": True
+            }
         })
 
     def test_failed_customer_creation_calls_form_invalid(self):
--- a/donate/payments/views.py
+++ b/donate/payments/views.py
@@ -593,7 +593,7 @@ class CardUpsellView(TransactionRequiredMixin, BraintreePaymentMixin, FormView):
         currency = form.cleaned_data['currency']
 
         # Create a subcription against the payment method
-        start_date = now().date() + relativedelta(months=1)     # Start one month from today
+        start_date = now().date() + relativedelta(months=1)     # Start recurring payment one month from today
         result = gateway.subscription.create({
             'plan_id': get_plan_id(currency),
             'merchant_account_id': get_merchant_account_id_for_card(currency),
@@ -700,7 +700,7 @@ class PaypalUpsellView(TransactionRequiredMixin, BraintreePaymentMixin, FormView
             return self.process_braintree_error_result(result, form)
 
         # Create a subscription against the payment method
-        start_date = now().date() + relativedelta(months=1)     # Start one month from today
+        start_date = now().date() + relativedelta(months=1)     # Start recurring payment one month from today
         result = gateway.subscription.create({
             'plan_id': get_plan_id(self.currency),
             'merchant_account_id': get_merchant_account_id_for_paypal(
--- a/donate/payments/views.py
+++ b/donate/payments/views.py
@@ -309,7 +309,9 @@ class CardPaymentView(BraintreePaymentMixin, FormView):
             'merchant_account_id': get_merchant_account_id_for_card(self.currency),
             'payment_method_token': payment_method.token,
             'price': form.cleaned_data['amount'],
-            'first_billing_date': now().date(),
+            'options': {
+                "start_immediately": True
+            }
         })
 
         if result.is_success:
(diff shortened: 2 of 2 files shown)


donate/payments/tests/test_views.py
```diff
@@ -450,7 +449,9 @@ def test_subscription_data_submitted_to_braintree(self):
             'merchant_account_id': 'usd-ac',
             'payment_method_token': 'payment-method-1',
             'price': 50,
-            'first_billing_date': date(2019, 7, 26),
+            'options': {
```
danielfmiranda, line 452: Based on this message from Braintree support:

"I’m Shaquille with Braintree Technical Support, happy to help.
I noted in the subscription examples provided the subscription was created with a specified start date in the future. This creates a subscription that remains Pending until the first billing date when the customer is charged and the subscription becomes Active. Do note, you can [create a subscription to start billing immediately](https://developer.paypal.com/braintree/docs/reference/request/subscription/create#set-start-date). This revision can be made to the plan that subscriptions are based on, or in the subscription.create() call itself."

It sounds like we are given two options when creating a new subscription: 
- Set the date and time for the subscription ourselves, which is what we were originally doing
- Or set the subscription to start immediately, which is what this PR is introducing. 

The reason for the change being that if a user submits a recurring donation past a certain time (around 7:30PM EST according to Jackie Lu) the user experiences a delay in recieving a donation receipt. 

Will Easton has also confirmed that when the user makes a donation at this time, the donation is set to "Pending" in braintree until the following day.

We believe this is due to differences in time zone between our application and Braintree. 

I believe that setting the donation to begin immediately will resolve this problem as it removes timezones from the equation.

donate/payments/views.py
```diff
@@ -309,7 +309,9 @@ def process_monthly_transaction(self, form):
             'merchant_account_id': get_merchant_account_id_for_card(self.currency),
             'payment_method_token': payment_method.token,
             'price': form.cleaned_data['amount'],
-            'first_billing_date': now().date(),
+            'options': {
```
danielfmiranda, line 312: Based on this message from Braintree support:

"I’m Shaquille with Braintree Technical Support, happy to help.
I noted in the subscription examples provided the subscription was created with a specified start date in the future. This creates a subscription that remains Pending until the first billing date when the customer is charged and the subscription becomes Active. Do note, you can [create a subscription to start billing immediately](https://developer.paypal.com/braintree/docs/reference/request/subscription/create#set-start-date). This revision can be made to the plan that subscriptions are based on, or in the subscription.create() call itself."

It sounds like we are given two options when creating a new subscription: 
- Set the date and time for the subscription ourselves, which is what we were originally doing
- Or set the subscription to start immediately, which is what this PR is introducing. 

The reason for the change being that if a user submits a recurring donation past a certain time (around 7:30PM EST according to Jackie Lu) the user experiences a delay in recieving a donation receipt. 

Will Easton has also confirmed that when the user makes a donation at this time, the donation is set to "Pending" in braintree until the following day.

We believe this is due to differences in time zone between our application and Braintree. 

I believe that setting the donation to begin immediately will resolve this problem as it removes timezones from the equation.

benhohner: Great description, thanks!

donate/payments/views.py
```diff
@@ -438,7 +440,9 @@ def form_valid(self, form, send_data_to_basket=True):
                 ),
                 'payment_method_token': payment_method.token,
                 'price': form.cleaned_data['amount'],
-                'first_billing_date': now().date(),
+                'options': {
+                    "start_immediately": True
```
benhohner, line 444:
```python
                    "start_immediately": True
```
Does this boolean ever end up in the Acoustic send mail function? I recall we had an issue with the serialization of Booleans before when sending a receipt email.

benhohner (approved):
R+

donate/payments/views.py
danielfmiranda: Hi Ben! This boolean does not make its way to that function, this one is just for creating their subscription in Braintree

donate/payments/views.py
WillatMozFdn: @danielfmiranda please make sure that this change applies ONLY to recurring donations that are made directly from the donate page, and NOT those that are done via the pop-up modal that asks for a recurring donation (which has to start in one month) after you give a one-time gift of (I think) $15 or more.

WillatMozFdn: (@danielfmiranda if this doesn't make sense, please ping me on slack.)

danielfmiranda: Hi @WillatMozFdn! 

I can confirm that this change does not make any changes to the donations made in the pop-up modal that appears after making a one time donation. 

Here is a link to the ["Upsell-View" ](https://github.com/mozilla/donate-wagtail/blob/29a0ec9fdb4e45d4204d5d38175044c1897de961/donate/payments/views.py#L596)which is displayed after a user has made a one time donation. we are still setting the next recurring donation to one month from the day the user made the initial one-time donation, should they choose to create a recurring one. 

Please let me know if you have any questions, thank you!

WillatMozFdn: Perfect, Daniel, thank you!  I had just wanted to triple-check on that ... because if we set ALL the recurrings to start immediately (both modal and regular), we'd have a whole separate set of donor care problems on our hands.  :-)  Thanks so much!!

PR #1515, opened 2021-06-29
1491 - Delete nav hamburger & most footer links from donate page

### LINKED ISSUE #1491: Delete nav hamburger & most footer links from donate page
We need to delete the hamburgler and most foot links from the donate page. 

See this [Airtable base for more information](https://airtable.com/tbljPmOCkK9JShaTS/viw9sdUJiIjCYpn0Y/rec9lfWpcwgenu0ZA?blocks=hide)

## Acceptance criteria 
- [x] Determine if those components should be deleted from _every_ page or just the main long-form page (page 2 of the donate process) 
- [x] Delete hamburger menu 
- [x] Remove most footer links

fessehaye:
Closes #1491

### DIFF
--- a/donate/templates/pages/base_page_master.html
+++ b/donate/templates/pages/base_page_master.html
@@ -22,39 +22,12 @@
 {% endblock body_top %}
 
 {% block navigation %}
-    <nav class="primary-nav" role="navigation" aria-label="Primary" data-primary-nav>
-        <ul class="primary-nav__container" role="menubar">
-            <li class="nav-item nav-item--primary" role="presentation">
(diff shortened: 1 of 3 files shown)


KalobTaulien:
If we're just deleting code, this looks good. But I want to check with Will first to see if we want to delete this code entirely, or just disable it. We might want to run an A/B test on this kind of thing in the future.

KalobTaulien (dismissed):
Slight change of plans. We need to hide the code, not delete it. 

A quick win would be to simply `{% comment %} .. {% endcomment %}` the sections we want to hide. 

A proper solution would be to add an option to the pages so editors can manage this. I think we just go for the quick win this time. But lets also make sure we have comments in there detailing why its all commented out instead of being deleted while using version control.

KalobTaulien (approved):

Issue #1518, opened 2021-07-07
Some countries without Post-Code still have Post-Code set as required.


danielfmiranda:
**Describe the bug**
Based on feedback from a user, they have attempted to make a donation with Hong Kong set as the country, which removes the post code field from view and should make the field not required so the user can submit their donation.

However, if you fill out the form and attempt to make a donation, you will recieve an error message, and when changing to country with post code, will see that it is still expecting a post code. 

**To Reproduce**
1. Go to the donation homepage
2. Fill out the form
3. Submit the form

**Expected behavior**
It should allow the user to submit their donation without a post code

danielfmiranda:
Interestingly, testing this on my local machine works as expected without a bug. 

However, I am able to recreate the bug on staging. 

I believe that the bug is in the back-end of the application, as all the front end is doing in regards to post code is removing the input from view. The back end is the portion that does the validation and will return a message if post-code is necessary.

I wonder if something when deployed is preventing the list from being loaded, thus making post code required for all countries?

danielfmiranda:
Hi Everyone, here are some findings after troubleshooting this issue:

- I can confirm that this issue is only reproducible in staging or prod. Whenever I try to reproduce this issue in my local machine, the error does not occur.
- I have added some print statements to see if the file "donate/payments/forms.py" was able to load the file "post-codes-list.json" and it appears that the error is caused by the python file having trouble finding and loading it. 
- As an attempt to fix this, I have tried the following paths to the file:
1. ".**/source/js/components/post-codes-list.json**" - which works on my local machine
2. "**../source/js/components/post-codes-list.json**" - does not work on local machine or heroku review apps as this is too far back in the directory. 
3. "**source/js/components/post-codes-list.json**" - Works again on local machine, not on heroku.

@Pomax I was wondering if you knew if there was something that is heroku specific that can be preventing the python file from reaching the post-codes-list information? Thanks in advance!

Pomax:
If I repeat the STR on staging, I see the dev console throw an error about `https://www.google-analytics.com/j/connect`, and looking at the CSP_CONNECT_SRC, there is `https://www.google-analytics.com/connect`, so let's set that to include that `/j/` and see what that does.

Pomax:
That fixed the CSP warning, but not the original error.

Pomax:
This works perfectly fine locally, so let's see if the DOM/source can tell us anything.

Pomax:
[...]

PR #1530, opened 2021-07-22
WIP: added exception printing to see if there is something not loading on …

danielfmiranda:
…review apps or staging

Closes #
Related PRs/issues #

## Checklist

_Remove unnecessary checks_

**Tests**
- [ ] Is the code I'm adding covered by tests?

**Changes in Models:**
- [ ] Did I update or add new fake data?
- [ ] Did I squash my migration?
- [ ] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/master/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

**Documentation:**
- [ ] Is my code documented?
- [ ] Did I update the documentation?

PR #1535, opened 2021-07-29
error logging for postal codes data

Pomax:
Investigatory PR for https://github.com/mozilla/donate-wagtail/issues/1518

### DIFF
--- a/donate/payments/forms.py
+++ b/donate/payments/forms.py
@@ -18,7 +18,7 @@ import json
 # # Loading a JSON list of countries and their respective post code formats if applicable,
 # # from the source/js directory.
 try:
-    with open('../source/js/components/post-codes-list.json') as post_code_data:
+    with open('/app/source/js/components/post-codes-list.json') as post_code_data:
         COUNTRY_POST_CODES = json.load(post_code_data)
 except Exception as error:
     print('ERROR: could not read in post codes list')
--- a/donate/payments/forms.py
+++ b/donate/payments/forms.py
@@ -18,7 +18,7 @@ import json
 # # Loading a JSON list of countries and their respective post code formats if applicable,
 # # from the source/js directory.
 try:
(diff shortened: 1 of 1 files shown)


danielfmiranda:
Thank you for the write up in the original ticket, and for creating this investigatory PR @Pomax! 

I had made [this PR](https://github.com/mozilla/donate-wagtail/pull/1530) in which I had tried to do the same thing (add an exception statement and see if heroku returns anything helpful), which I will now close and add my findings to this ticket instead.

In the PR linked above, I added an exception statement to see if the .JSON file that houses the data for countries was even being loaded, and can confirm that it returned a message stating that it was not able to load the JSON file.

This led me to believe it was a path problem, so I had tried a couple of different paths to the file to see if they would work, which all returned the same "Not able to load the file" error message. 

Here is a list of the different paths that I have tried: 

- ./source/js/components/post-codes-list.json (The original path, which did not work)
- source/js/components/post-codes-list.json (I thought starting from root directory like this would work, but recieved same error message
- ..source/js/components/post-codes-list.json (I tried 2 dots in order to leave the python directory that the forms.py file was housed in, however this still returned an error.)

After the last try

danielfmiranda:
Upon further investigation, I found that the heroku logs indeed show us that the path to the json file was not correct, preventing it from loading in heroku. 

heroku_logs_post_code_error

However, I noted that 2 or 3 lines above that error statement, it also threw an error about looking for a .env file which is housed in the same level as the source directory. It appears that heroku is trying to access the .env file through "/app/..."

heroku_logs_post_code_app_dir

I then made an update to our python file, to change the directory to "/app/source/js/components/post-codes-list.json" and can confirm that I no longer see an error message in the logs, and the donate page is now allowing me to make a donation using Hong Kong as a country, with no ZIP code.

danielfmiranda:
Closing as we now have #1543 with the suggested fix for this issue.

PR #1560, opened 2021-09-29
added updated list of languages

### LINKED ISSUE #1557: Update donor support forms to reflect more language picklist options
In both Mozilla and Thunderbird donate support forms, update the HTML to support the expanded list of languages:

@stephaniemcv can you please paste the updated HTML into a comment in this issue.

danielfmiranda:
Closes #1557 
Related PRs/issues #

## Checklist

**Changes in Models:**

### DIFF
--- a/donate/templates/pages/core/contributor_support_page_master.html
+++ b/donate/templates/pages/core/contributor_support_page_master.html
@@ -87,13 +87,20 @@
                         <div class="form-item__wrapper form-custom-select">
                             <select id="00N4x00000O6Pzw" name="00N4x00000O6Pzw" title="Language">
                                 <option value="">--{% trans "None" context "Option in 'Language' dropdown" %}--</option>
-                                <option value="English">{% trans "English" %}</option>
-                                <option value="French">{% trans "French" %}</option>
-                                <option value="German">{% trans "German" %}</option>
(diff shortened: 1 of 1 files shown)


danielfmiranda:
Hi @TheoChevalier! 

I tagged you in this issue because I noticed that while all of the new options added are going to be translated, it appears that only the "None" and "Other"choices have a context set.

I was wondering if there was a reason for this, or if you would like me to add context for every choice. 
Thanks!

TheoChevalier (changes requested):
Hey Daniel,

No need to add specific context to the language names, but it’s been done for "None" and "Other" for specific reasons, yes:
- Making sure these translations are only used here, as these strings exist in other (different) contexts
- Related to the above, help translators understand what they refer to, so they can adapt the translation (declension, genitive case, plural, etc). For instance in French, the word language  (langue) is feminine, so "None" must use the feminine form too, "Aucune".

I’ve filed #1561 as a follow up to properly sort this list in all languages as this becomes a problem with a longer list

danielfmiranda:
Thanks for the catches @TheoChevalier! I have made the desired updates and re-requested review 😁

TheoChevalier (approved):
Thanks! Merging & getting translation started

PR #1562, opened 2021-09-29
uplift wagtail to 2.14.1

### LINKED ISSUE #1559: Update wagtail to version 2.14
Update the wagtail dependency to 2.14, this also requires updating any other packages to ensure they are compatible with the new version of wagtail.

Pomax:
Closes #1559 

This also includes a switch to a direct git URL for the `wagtail-ab-testing` depedency, as its code was updated to support wagtail 2.14 but no version on pypi includes that change yet.

As far as I can tell this passes all tests, and the pages I checked looked normal.


danielfmiranda (approved):
Hi Pomax, thanks for tagging me in this! The code looks good and I can confirm that the site still works with the new/updated versions. 

Approved!

TheoChevalier (changes requested):
Let’s bump wagtail-localize to `1.0rc4` too, because RC3 added support for 2.14, and RC2 for 2.13 (it didn’t complain because it’s currently unpinned)

Pomax:
With https://github.com/torchbox/wagtail-ab-testing/issues/40 landed, we can remove the github dependency for wagtail-ab-testing

Pomax:
@TheoChevalier I've updated both libs, if this looks good to you we can merge it in =D

PR #1567, opened 2021-11-03
Add description to RQ job for braintree webhook

### LINKED ISSUE #1564: Add `description` to RQ queueing instructions
RQ will logs the job `description` on success, which if unset seems to default to the job arguments, meaning it will log a huge braintree string on success. We should be able to pass a `description` kwarg to our `queue.enqueue` calls, to make sure that we only log short and sweet text.

### LINKED ISSUE #1566: Braintree Signatures Polluting Logs
**Describe the bug**
Braintree signatures are being logged.
For some reason the string:
`donate.payments.tasks.process_webhook({'bt_signature': <massive amounts of text>})`
is being logged which is consuming our logstash bandwidth and polluting logs, also possibly polluting logs with PII.

**To Reproduce**
1. Go to Logstash
2. Search for bt_signature

**Expected behavior**
User information derived signatures shouldn't be logged. 
Let's determine why this was implemented in the first place, and if we can replace it with a better message or get rid of the statement all together.

tomusher:
Closes #1564 by adding a description to the task enqueued by the Braintree webhook, overriding the default behaviour of using the entire payload as a description. Closes #1566


Pomax (approved):

Pomax (changes requested):
This seems to be breaking CI

Pomax (approved):

PR #1586, opened 2021-11-30
small fixes

Pomax:
No associated issue, addresses some errors flagged on the donate Sentry dashboard.

### DIFF
--- a/donate/templates/payment/includes/trigger_ab_testing_thank_you_event.html
+++ b/donate/templates/payment/includes/trigger_ab_testing_thank_you_event.html
@@ -1,10 +1,13 @@
 <script>
-    document.addEventListener('DOMContentLoaded', function() {
-        // Need to setTimeout to make sure the A/B testing library is loaded
-        setTimeout(function() {
-            if (window.wagtailAbTesting) {
+    document.addEventListener('DOMContentLoaded', function () {
+        // Try to send an A/B signal to the wagtail A/B testing framework
+        (function tryWagtailABEvent(count) {
+            if (count > 5) return;
+            if (window.wagtailAbTesting && window.wagtailAbTesting.triggerEvent) {
                 window.wagtailAbTesting.triggerEvent('visit-thank-you-page');
+            } else {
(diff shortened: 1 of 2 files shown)


donate/templates/payment/includes/trigger_ab_testing_thank_you_event.html
```diff
@@ -1,10 +1,13 @@
 <script>
-    document.addEventListener('DOMContentLoaded', function() {
-        // Need to setTimeout to make sure the A/B testing library is loaded
-        setTimeout(function() {
-            if (window.wagtailAbTesting) {
+    document.addEventListener('DOMContentLoaded', function () {
+        // Try to send an A/B signal to the wagtail A/B testing framework
+        (function tryWagtailABEvent(count) {
+            if (count > 5) return;
+            if (window.wagtailAbTesting && window.wagtailAbTesting.triggerEvent) {
```
Pomax, line 6: the error in sentry was about the fact that `window.wagtailAbTesting.triggerEvent` did not exist even though `window.wagtailAbTesting` did, so this updates the code to make sure to test for the function before calling it, while also running up to five attempts before giving up.

source/js/payments-card.js
```diff
@@ -21,13 +21,26 @@ function setupBraintree() {
     );
 
   function showErrorMessage(msg) {
-    errorDiv.toggleAttribute("hidden", false);
-    errorDiv.innerHTML = msg;
+    if (errorDiv) {
```
Pomax, line 24: the error in Sentry was about the fact that we called `errorDiv.toggleAttribute("hidden", false);` while `errorDiv` was undefined. So this update makes sure to check whether it exists, and if not, put the error data on the console so it's at least "somewhere".

danielfmiranda (approved):
Hi Pomax, thanks for taking the time to update these files as if I understand correctly, these errors have been making it more difficult to trouble shoot than usual for multiple people on our team. 

I think updating this will decrease our trouble shooting times by making the actual errors easier to see.

Approved!

Issue #1585, opened 2021-11-30
help form recaptcha is not working correctly


Pomax:
There's some very odd code in the `contributor_support_page_master.html` template that looks like this:

```
    <script src="https://www.google.com/recaptcha/api.js"></script>
    <script>
        function timestamp() {
            var response = document.getElementById("g-recaptcha-response");
            if (response == null || response.value.trim() == "") {
                var elems = JSON.parse(document.getElementsByName("captcha_settings")[0].value);
                elems["ts"] = JSON.stringify(new Date().getTime());
                document.getElementsByName("captcha_settings")[0].value = JSON.stringify(elems);
            }
        }
        setInterval(timestamp, 500);
```

And (a) why is this code needed, (b) why does this run forever, and (c) that `document.getElementsByName("captcha_settings")[0]` may not exist yet when this code runs.

It feels like this was copy-pasted from somewhere instead of implemented based on google's "here's how you embed the recaptcha" instructions?

PR #1587, opened 2021-11-30
fix the recaptcha for help

Pomax:
closes https://github.com/mozilla/donate-wagtail/issues/1585

You can test this by creating a "help" page in the CMS and then visiting that page - you should see it add the recaptcha but leave the submit button disabled until the recaptcha's been clicked.

contrast this to the help page on https://donate.mozilla.org/en-US/help/ which will technically let you submit without checking the recaptcha (although it will then get silently rejected by our post handler)

### DIFF
--- a/donate/templates/pages/core/contributor_support_page_master.html
+++ b/donate/templates/pages/core/contributor_support_page_master.html
@@ -7,18 +7,24 @@
 {% endblock %}
 
 {% block head_addenda %}
-    <script src="https://www.google.com/recaptcha/api.js"></script>
-    <script>
-        function timestamp() {
-            var response = document.getElementById("g-recaptcha-response");
-            if (response == null || response.value.trim() == "") {
-                var elems = JSON.parse(document.getElementsByName("captcha_settings")[0].value);
-                elems["ts"] = JSON.stringify(new Date().getTime());
-                document.getElementsByName("captcha_settings")[0].value = JSON.stringify(elems);
-            }
+    <script type="text/javascript">
+    function hookupFormSubmitButton() {
+        try {
+            const recaptcha = document.getElementById("g-recaptcha");
+            console.log(recaptcha);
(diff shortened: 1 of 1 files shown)


donate/templates/pages/core/contributor_support_page_master.html
```diff
@@ -7,18 +7,24 @@
...
+                    submitButton.removeAttribute("disabled");
+                }
+            });
+        } catch (e) {
+            console.error(e);
         }
-        setInterval(timestamp, 500);
+      };
     </script>
+    <script src="https://www.google.com/recaptcha/api.js?onload=hookupFormSubmitButton&render=explicit" async defer></script>
```
Pomax, line 27: this switches the recaptcha over to running the `hookupFormSubmitButton` function after it boostraps itself

donate/templates/pages/core/contributor_support_page_master.html
```diff
@@ -7,18 +7,24 @@
...
+    <script type="text/javascript">
+    function hookupFormSubmitButton() {
+        try {
+            const recaptcha = document.getElementById("g-recaptcha");
+            console.log(recaptcha);
+            grecaptcha.render(recaptcha, {
+                sitekey: "{{ help_recaptcha_site_key }}",
+                callback: (token) => {
+                    var submitButton = document.querySelector('.support-form input[type="submit"]');
+                    submitButton.removeAttribute("disabled");
```
Pomax, line 19: note that we're not _verifying_ this recaptcha token atm, we're simply using it to unlock the submit button. That's not "secure" (an issue has been filed to determine whether we need to do server validation before proxying the data on to our help endpoint).

danielfmiranda (approved):
Hi Pomax! Just took a look and it appears that everything is working as expected, I did want to note however that its giving me a salesforce related error when clicking submit, which I am assuming is related to the page being localhost. 
image

Issue #1596, opened 2021-12-06
Add integration testing


Pomax:
We only have Python-level unit testing at the moment, which means we have nothing in place to test things like "do recaptchas load", "are there client-side errors on any of the pages we need", "does page navigation work", etc. and it's hit us _hard_ in the last few weeks.

I would recommend we add `playwright` to this repo with a slew of integration tests to make sure all critical page flows are covered.

PR #1604, opened 2021-12-16
Add a request session variable that regulates when a user can load the CC donation page

Pomax:
RFC PR to see if this is worth actually putting in place.

This code change introduces a wagtail hook that "kicks in" when users load any wagtail-managed page, such as our landing page or specific campaign pages, adding a `may_load_card_page` flag to their session. 

The card view, which is _not_ managed by wagtail, then checks for that flag on load (on all HTTP verbs, meaning for our purposes: both standard page GET as well as form POST). If the flag has not been set, the user navigated to the card page URL without ever going through a "pick your payment option", which means something fishy is going on, and they get redirected to the landing page instead.

This includes if someone visits the landing page, clicks through to the credit card page, and then with their flag set, loads up 100+ credit card page tabs. Once any of those tabs is used to submit a CC donation, any subsequent attempts to post will simply get thrown away, with the browser being sent back to the landing page instead.

## Additional acceptance criteria for this PR

- [ ] tie this functionality to a feature flag 
- [ ] add integration tests once #1596 has been resolved.


jzinner (approved):
So I reviewed these changes and I reviewed the functionality using the testing app.  

I guess this would stop a bot attack - since a bot wouldn't likely preserve the session variables?  Or have a session.

Approving

mtdenton:
Any idea why this PR never got merged?

Pomax:
It had two reviewer requests, only one came in, and this was a fairly loaded problem to tackle at the time so we didn't want to jump the gun and land something before it was fully reviewed. I don't know if that's still the case.

PR #1625, opened 2022-01-18
Add integration testing using playwright

### LINKED ISSUE #1596: Add integration testing

Pomax:
This adds the same kind of integration testing that we have on the foundation repo to the donate repo, so we can do actual page testing.

Note that the fact that this runs in CI and passes is pretty much "this works", but to run this locally the instructions are basically the same as for the foundation repo. 

Closes #1596

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -61,3 +61,54 @@ jobs:
         flake8 donate/
         coverage run --source './donate' ./manage.py test --settings=donate.settings --configuration=Testing
 
+  test_integration:
+    name: Integration testing
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:10
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: donate
+        ports:
+        - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+    env:
+      # Update these for tests on github! 
+      ALLOWED_HOSTS: localhost
+      CONTENT_TYPE_NO_SNIFF: True
+      DATABASE_URL: postgres://postgres:postgres@localhost:5432/donate
+      DEBUG: False
+      DJANGO_SECRET_KEY: secret
+      SET_HSTS: False
+      SLACK_WEBHOOK_PONTOON: ${{ secrets.SLACK_WEBHOOK_PONTOON }}
+      SSL_REDIRECT: False
+      USE_S3: False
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -46,10 +49,10 @@
     "babel-polyfill": "^6.26.0",
     "braintree-web": "^3.62.2",
     "intersection-observer": "^0.10.0",
-    "node-sass": "^4.14.1",
```
Pomax, line 49: node-sass is both obsolete and incompatible with modern Node. Replaced with the proper `sass` package

tests/wait-for-images.js
```diff
@@ -0,0 +1,41 @@
+/**
+ * Playwright has no built-in way to wait for all on-page images
```
Pomax, line 2: same file as on the foundation repo

.github/workflows/continuous-integration.yml
```diff
@@ -61,3 +61,54 @@ jobs:
...
+      X_FRAME_OPTIONS: DENY
+      XSS_PROTECTION: True
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 14
```
danielfmiranda, line 97:
```suggestion
        node-version: 14
        cache: 'npm'
```

.github/workflows/continuous-integration.yml
```diff
@@ -61,3 +61,54 @@ jobs:
...
+      SLACK_WEBHOOK_PONTOON: ${{ secrets.SLACK_WEBHOOK_PONTOON }}
+      SSL_REDIRECT: False
+      USE_S3: False
+      X_FRAME_OPTIONS: DENY
+      XSS_PROTECTION: True
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
```
danielfmiranda, line 94:
```suggestion
        python-version: 3.9
        cache: 'pip'
```

danielfmiranda (approved):
Hi Pomax, this looks good to me so I am approving! 👍 

However I did add two suggested comments just in case you wanted to cache the CI runs like we do in https://github.com/mozilla/foundation.mozilla.org/pull/8100

Pomax:
good call

PR #1630, opened 2022-01-25
form no longer submits if required fields are left out

### LINKED ISSUE #1395: Consider adding JavaScript validation to the second donate page.
On the second page when filling out your personal information, if the email address doesn't validate it refreshes the page and loses credit card info. 

We shouldnt store the credit card info, but we can add JavaScript validation to the page so it doesn't refresh and the user doesn't need to re-enter their information. 

Having users re-enter their information, especially a credit card number which isn't memorized, could lead to donation dropoffs. 

We should look into adding JavaScript form validation before Django form validation kicks in for a better user experience. 

Steps to reproduce: 
1. Go to: https://donate.mozilla.org/en-CA/card/single/?source_page_id=415&currency=cad&amount=20
2. Fill out fake data, but don't use a real email address. Use `test` with no `@` symbol in it. 
3. Submit the form. 
4. Page should refresh with missing data causing the user to re-enter everything.

danielfmiranda:
Closes #1395

Link to test page:
https://donate-wagta-frontend-v-c5ymdv.herokuapp.com/en-US/card/single/?source_page_id=3&currency=usd&amount=20

**Steps to test**:

1. Visit the test page above, and fill out the form, leaving out whichever fields in the "Your details" section that you would like to test.
2. Fill out the credit card information with this test information:
Credit Card #: 4111 1111 1111 1111
EXP: 12/23
CVV: 123

3.) Click the submit button
4.) If anything was incorrect or left blank, the fields should be highlighted red, and there should be an error message at the top. 
5.) Correct these fields and then submit
6.) Everything should go through as expected! If so, testing is complete

### DIFF
--- a/donate/templates/payment/card_master.html
+++ b/donate/templates/payment/card_master.html
@@ -55,10 +55,6 @@
                                 {% for error in form.non_field_errors %}
                                 <div class="form__errors">{{ error }}</div>
                                 {% endfor %}
-                            {% elif form.errors %}
-                                <div class="form__errors">
-                                    {% trans "There were some errors with your form. Please amend the fields highlighted below." %}
-                                </div>
                             {% elif gateway_address_errors %}
                                 <div class="form__errors">
                                     <p>{% trans "We were unable to take payment because the payment processor reported an error with your address information:" %}</p>
@@ -67,15 +63,19 @@
                                     {% endfor %}
                                     {% trans "Please correct these errors and try again." %}
                                 </div>
+                            {% else %}
+                                <div id="your-details-form-error" class="form__errors hidden">
+                                    {% trans "There were some errors with your form. Please amend the fields highlighted below." %}
+                                </div>
                             {% endif %}
 
-                            {% render_form_field form.email %}
-                            {% render_form_field form.first_name %}
-                            {% render_form_field form.last_name %}
-                            {% render_form_field form.address_line_1 %}
+                            {% render_form_field form.email front_end_validated=True %}
+                            {% render_form_field form.first_name front_end_validated=True %}
+                            {% render_form_field form.last_name front_end_validated=True %}
+                            {% render_form_field form.address_line_1 front_end_validated=True %}
                             <div class="form__group form__group--city-post">
-                                {% render_form_field form.post_code css_classes="post-code-container" %}
-                                {% render_form_field form.city %}
+                                {% render_form_field form.post_code css_classes="post-code-container" front_end_validated=True %}
+                                {% render_form_field form.city front_end_validated=True %}
                             </div>
                             {% render_form_field form.country %}
 
--- a/source/js/payments-card.js
+++ b/source/js/payments-card.js
@@ -7,7 +7,7 @@ import gaEvent from "./components/analytics";
 function setupBraintree() {
   var paymentForm = document.getElementById("payments__braintree-form"),
     nonceInput = document.getElementById("id_braintree_nonce"),
-    postCodeInputContainer = document.querySelector('.post-code-container'),
+    postCodeInputContainer = document.querySelector(".post-code-container"),
     deviceDataInput = document.getElementById("id_device_data"),
     captchaInput = document.getElementById("id_captcha"),
     captchaEnabled = captchaInput !== null,
@@ -82,7 +82,6 @@ function setupBraintree() {
   }
 
   function checkYourDetailsFields() {
-    console.log(postCodeInputContainer)
     // Looping through required "Your Details" fields, in order to prevent a bad form submit,
     // which will result in CC data loss.
 
--- a/source/js/payments-card.js
+++ b/source/js/payments-card.js
@@ -74,11 +74,10 @@ function setupBraintree() {
   }
 
   function removeDetailFieldError(field) {
-      field.parentNode.classList.remove("form-item--errors");
-      document
-        .getElementById(`error-message__` + `${field.name}`)
-        .classList.add("hidden");
-    
+    field.parentNode.classList.remove("form-item--errors");
+    document
+      .getElementById(`error-message__` + `${field.name}`)
+      .classList.add("hidden");
   }
 
   function checkYourDetailsFields() {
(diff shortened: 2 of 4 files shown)


donate/templates/payment/card_master.html
```diff
@@ -67,15 +63,19 @@ <h1 class="heading heading--primary heading--bottom-margin">{% trans "Your Detai
                                     {% endfor %}
                                     {% trans "Please correct these errors and try again." %}
                                 </div>
+                            {% else %}
```
danielfmiranda, line 66: What this section is now doing is saying: 
"If there are any errors on the "your details" form, display this translated message instead of refreshing the page."

However, if the form is completed and submitted, but there is an error on the braintree side or a network error, the page will refresh and will render the appropriate error

source/js/payments-card.js
```diff
@@ -15,11 +15,23 @@ function setupBraintree() {
       "An error occurred. Please reload the page or try again later."
     ),
     errorDiv = document.getElementById("payments__braintree-errors-card"),
+    yourDetailsFormErrorDiv = document.getElementById(
```
danielfmiranda, line 18: Div at the top of the form that will render saying "there were some errors with your form. Please correct the highlighted fields" if necessary

source/js/payments-card.js
```diff
@@ -15,11 +15,23 @@ function setupBraintree() {
...
+      "your-details-form-error"
+    ),
     braintreeErrorClass = "braintree-hosted-fields-invalid",
     braintreeParams = JSON.parse(
       document.getElementById("payments__braintree-params").textContent
     );
 
+  var requiredFields = {
```
danielfmiranda, line 26: required fields for the "your details" form

source/js/payments-card.js
```diff
@@ -51,6 +63,50 @@ function setupBraintree() {
     container.classList.remove(braintreeErrorClass);
   }
 
+  function showDetailFieldError(field) {
```
danielfmiranda, line 66: Function to show error state on any field that needs it

source/js/payments-card.js
```diff
@@ -51,6 +63,50 @@ function setupBraintree() {
...
+    // If field error is already showing, do not render again
+    if (!field.parentNode.classList.contains("form-item--errors")) {
+      field.parentNode.classList.add("form-item--errors");
+      document
+        .getElementById(`error-message__` + `${field.name}`)
+        .classList.remove("hidden");
+    }
+  }
+
+  function checkYourDetailsFields() {
```
danielfmiranda, line 76: Function that loops through all required fields and checks them. If any are not filled out, it will then display the error state.

source/js/payments-card.js
```diff
@@ -51,6 +63,50 @@ function setupBraintree() {
...
+    // which will result in CC data loss.
+
+    var yourDetailsFormValidity = true;
+
+    for (const [fieldKey, inputElement] of Object.entries(requiredFields)) {
+      // Since postcode is not always rendered and needed, we need a special case for it.
+      if (fieldKey == "postCodeInput") {
+        // If Post Code Field is rendered, verify its filled out.
+        if (
+          !inputElement.parentNode.parentNode.classList.contains("hidden") &
```
danielfmiranda, line 87: Hi @fessehaye! I was wondering if you knew any other way to preform this line as "parentNode.parentNode" does not feel like the best way to do this, but it works!

Pretty much what I am trying to do is see if the div that houses the "postal code" input element has the "hidden" class, however it is 2 levels up

Thanks in advance!

source/js/payments-card.js
```diff
@@ -117,13 +173,20 @@ function setupBraintree() {
 
         submitButton.addEventListener("click", function (e) {
           // Trigger browser form validation
+          e.preventDefault();
+
+          if (!checkYourDetailsFields()) {
```
danielfmiranda, line 178: If "your details" form is invalid, prevent form from being submitted!

sabrinang (changes requested):
The error fields are working to prevent me from submission. 

A use case I tried is:
- I left out two fields clicked donate and it then highlighted two required fields
- I go back and fill out one and click donate again but my filled out field remains highlighted as required
FireShot Capture 015 - Your information - Donate to Mozilla_ - donate-wagta-frontend-v-c5ymdv herokuapp com

What I expected was that when I clicked donate after filling out one and still having a required field is that the one I filled should no longer be highlighted as required like the other empty one. Is this fixable? Otherwise, it worked as expected.

danielfmiranda:
Hi @sabrinang! Thanks for reviewing, I have updated the form to now un-highlight error fields that have been filled out!

source/js/payments-card.js
```diff
@@ -51,6 +63,62 @@ function setupBraintree() {
...
+    // which will result in CC data loss.
+
+    var yourDetailsFormValidity = true;
+
+    for (const [fieldKey, inputElement] of Object.entries(requiredFields)) {
+      // Since postcode is not always rendered and needed, we need a special case for it.
+      // If postcode's parent div is rendered/required, and the input value is empty:
+      if (
+        fieldKey == "postCodeInput" &&
+        !inputElement.parentNode.parentNode.classList.contains("hidden") &&
```
danielfmiranda, line 94: Hi @fessehaye! I was wondering if you knew any other way to preform this line as "parentNode.parentNode" does not feel like the best way to do this, but it works!

Pretty much what I am trying to do is see if the div that houses the "postal code" input element has the "hidden" class, however it is 2 levels up

Thanks in advance!

sabrinang (approved):
The fields that were previously required now un-highlight when filled as expected 👍 thank you Daniel!

source/js/payments-card.js
fessehaye: I would almost consider just adding a CSS class to the parent and then `document.querySelector` that class

fessehaye (approved):

source/js/payments-card.js
danielfmiranda: Great idea Simon! It looks like since we had to do custom logic with post code before, it already had a custom class name that I can tap into. Updated it to use that!

PR #1637, opened 2022-02-02
minimum amount message now rendering and updating

### LINKED ISSUE #1634: Update message for "other amount" field
Donor Care is requesting that we show the minimum amount message on the "Other amount" field when the user clicks on the form input.

However, theres some discussion whether or not we should just show the minimum amount underneath the field constantly.

- [x] Reach out to design to see which format they think is best
- [ ] Implement suggestion

danielfmiranda:
Closes #1634

**Link to test page**: https://donate-wagta-other-amou-ejbati.herokuapp.com/en-US/

**Steps to test**:

1. Visit the test page linked above
2. Note the new "Minimum amount" message below the "other" value field.
Screen Shot 2022-02-03 at 3 24 14 PM

3. If you type in a number below the minimum amount and click away from the input, a error should pop up below.
Screen Shot 2022-02-03 at 3 25 26 PM

4. If you type a number above the minimum amount and click away from the input, the error should disappear.
5. Please select a new currency from the drop down at the top right.
Screen Shot 2022-02-03 at 3 24 14 PM

6. Once selected, the form should update, and the minimum amount message should also update with a new currency.
Screen Shot 2022-02-03 at 3 26 42 PM

7. Please try steps 3 & 4 again, and if they work, testing is complete!


donate/templates/fragments/donate_form.html
```diff
@@ -107,12 +110,15 @@ <h2 class="donate-form__actions-heading heading heading--secondary">{% trans "Ch
...
+                            <input type="radio" class="donation-amount__radio monthly-amount-donation-radio" name="amount" value="other" id="monthly-amount-other" autocomplete="off" data-other-amount-radio="">
                             <label for="monthly-amount-other" class="donation-amount__label donation-amount-other__label" data-currency="">{% get_localized_currency_symbol initial_currency_info.code %}</label>
                             <label for="monthly-amount-other-input" class="donation-amount__label--hidden">{% trans "Other amount" %}</label>
                             <input type="number" class="donation-amount__input monthly-amount-other-input" name="monthly-amount-other-input" id="monthly-amount-other-input" min="{{ initial_currency_info.minAmount }}" max="10000000" placeholder="{% trans 'Other amount' %}" data-other-amount="">
                         </div>
-                        <p class='error-message error-message__monthly-other-amount hidden'>
+                        <p class='minimum'>
+                            {% trans "Minimum amount is" %} {% get_localized_currency_symbol initial_currency_info.code %}{{ initial_currency_info.minAmount }}
```
TheoChevalier, line 119:
```suggestion
                            {% format_currency request.LANGUAGE_CODE initial_currency_info.code initial_currency_info.minAmount as minimum_amount %}
                            {% blocktrans with min_amount=minimum_amount trimmed %}Minimum amount is {{ min_amount }}{% endblocktrans %}
```

source/js/components/currency-selector.js
```diff
@@ -96,10 +99,26 @@ class CurrencySelect {
...
-      `<div class='donation-amount donation-amount--two-col donation-amount--other'><input type='radio' class='donation-amount__radio' name='amount' value='other' id='${type}-other' autocomplete='off' data-other-amount-radio><label for='${type}-other' class='donation-amount__label' data-currency>$</label><input type='number' class='donation-amount__input' id='${type}-other-input' placeholder='${otherAmountString}' data-other-amount min="${minAmount}" max="10000000"></div>`
+      `
+      <div class='donation-amount--two-col'>
+        <div class='donation-amount donation-amount--other'>
+          <input type='radio' class='donation-amount__radio ${type}-donation-radio' name='amount' value='other' id='${type}-other' autocomplete='off' data-other-amount-radio>
+          <label for='${type}-other' class='donation-amount__label' data-currency>$</label>
+          <input type='number' class='donation-amount__input' id='${type}-other-input' placeholder='${otherAmountString}' data-other-amount min="${minAmount}" max="10000000">
+        </div>
+        <p class='minimum'>
+          ${minimumOtherAmountString} ${symbol}${minAmount}
```
TheoChevalier, line 114: We will need to adapt the approach here as well, I think you can use the `formattedValue` variable directly (no need to pass the symbol explicitly), combined with [string interpolation](https://docs.djangoproject.com/en/4.0/topics/i18n/translation/#interpolate)?

TheoChevalier (changes requested):
Hey @danielfmiranda, I’ve left a few comments related to internationalization, would you mind having a look?

Thanks!

danielfmiranda:
Thank you for the feedback @TheoChevalier! I have implemented your requests and re-requested review.

source/js/components/currency-selector.js
danielfmiranda: Thanks for the suggestion Theo! using the formatter that was used to create formattedValue works great, now when the user selects a currency, the symbol matches the format on the buttons. Instead of directly injecting the symbol.

kristinashu (approved):
Design looks good!

fessehaye (commented):
Review app is down!

TheoChevalier (approved):
Thanks Daniel, the l10n part looks good now, just left a code-related comment

fessehaye (approved):

fessehaye (commented):
Looks good, only thing I would suggest is replacing some vars with `let/const` since it looks like you are jumping between them

PR #1644, opened 2022-02-24
added reworded error message

### LINKED ISSUE #1635: Paypal window does not appear, and donate page shows error message
## Tasks:

- [x] Find out how to reproduce error (see The Issue below) 
- [x] Figure out what is causing the error - outcome: Error caused by http 429 being returend from paypal
- [x] Update stakeholders (Stephanie Lowry) regarding source of error and possible fixes or improvements
- _Figure out how to catch http 429 status codes_ - the [PayPal Checkout.js library](https://github.com/paypal/paypal-checkout) function that is used [braintree-web/paypal-checkout](https://braintree.github.io/braintree-web/3.24.1/module-braintree-web_paypal-checkout.html#.create) does not  return the http status codes.  So this part of the ask is dropped.
- [ ] Change [generalErrorMsg in paypal.js ](https://github.com/mozilla/donate-wagtail/blob/50a365891aadc51bb85b226a410dafb0c398d4f9/source/js/components/paypal.js#L17) to read
`Unable to connect to Paypal.  Please try again in a few minutes. `

### The issue
the issue presents before the donation is made - so after entering a value into donate other amount field, which is under $10 then clicking the Paypal button - the Paypal popup window never appears and the error message: “There was an error processing your payment. Please try again.” 

Screen Shot 2022-01-19 at 10 58 23 AM

Screen Shot 2022-01-18 at 1 44 27 PM

Expected Result:
Screen Shot 2022-01-28 at 2 42 14 PM

  Resolution for this would be: better error message.  “Could not connect to paypal server.  Please disable pop-blockers or try a different browser.”  

###Steps:
- [x] Try to reproduce this by maybe by enabling a pop-up blocker on a browser.   If having trouble reach out to Stephanie on donor care to see how they reproduce the issue. 

Paypal Error Docs: https://developer.paypal.com/docs/checkout/standard/customize/handle-errors/

danielfmiranda:
Closes #1635

**Steps to test**:
Testing is not really possible as the review apps use paypal/braintree sandbox mode.

However, here is a screenshot of what appears after following the same workflow that we have used to reproduce the error on my local machine:
Screen Shot 2022-02-24 at 1 29 37 PM

I would also like to note that this new reworded error message only appears if there is a *general* error from paypal, which rate limiting is included in.

If the payment gets cancelled by the user, the "Payment cancelled" message still appears, and if there is an error authorizing the payment, the original "There was an error processing your payment. Please try again." error message gets rendered.

### DIFF
--- a/source/js/components/paypal.js
+++ b/source/js/components/paypal.js
@@ -17,7 +17,7 @@ export default function initPaypal(
     "There was an error processing your payment. Please try again."
   );
   var generalErrorMsg = window.gettext(
-    "Unable to connect to PayPal. Please try again in a few minutes."
+    "We were unable to establish a connection to PayPal. Please try again in a few minutes."
   );
   var braintreeParams = JSON.parse(
     document.getElementById("payments__braintree-params").textContent
--- a/source/js/components/paypal.js
+++ b/source/js/components/paypal.js
@@ -13,8 +13,11 @@ export default function initPaypal(
   var loadingErrorMsg = window.gettext(
     "An error occurred. Please reload the page or try again later."
   );
-  var generalErrorMsg = window.gettext(
+  var authErrorMsg = window.gettext(
     "There was an error processing your payment. Please try again."
+  );
+  var generalErrorMsg = window.gettext(
+    "Unable to connect to Paypal. Please try again in a few minutes."
   );
   var braintreeParams = JSON.parse(
     document.getElementById("payments__braintree-params").textContent
@@ -82,7 +85,7 @@ export default function initPaypal(
                   payload
                 ) {
                   if (err) {
-                    showErrorMessage(generalErrorMsg);
+                    showErrorMessage(authErrorMsg);
                     return;
                   }
 
--- a/source/js/components/paypal.js
+++ b/source/js/components/paypal.js
@@ -17,7 +17,7 @@ export default function initPaypal(
     "There was an error processing your payment. Please try again."
   );
   var generalErrorMsg = window.gettext(
-    "Unable to connect to Paypal. Please try again in a few minutes."
+    "Unable to connect to PayPal. Please try again in a few minutes."
   );
   var braintreeParams = JSON.parse(
     document.getElementById("payments__braintree-params").textContent


source/js/components/paypal.js
```diff
@@ -13,9 +13,12 @@ export default function initPaypal(
   var loadingErrorMsg = window.gettext(
     "An error occurred. Please reload the page or try again later."
   );
-  var generalErrorMsg = window.gettext(
+  var authErrorMsg = window.gettext(
     "There was an error processing your payment. Please try again."
   );
+  var generalErrorMsg = window.gettext(
+    "Unable to connect to Paypal. Please try again in a few minutes."
```
TheoChevalier, line 20: Tiny typo

```suggestion
    "Unable to connect to PayPal. Please try again in a few minutes."
```

danielfmiranda: Great catch @TheoChevalier! I have updated it to reflect your feedback

TheoChevalier (approved):
Thank you!

source/js/components/paypal.js
```diff
@@ -13,9 +13,12 @@ export default function initPaypal(
   var loadingErrorMsg = window.gettext(
     "An error occurred. Please reload the page or try again later."
   );
-  var generalErrorMsg = window.gettext(
+  var authErrorMsg = window.gettext(
     "There was an error processing your payment. Please try again."
   );
+  var generalErrorMsg = window.gettext(
+    "Unable to connect to PayPal. Please try again in a few minutes."
```
Pomax, line 20: this phrasing doesn't feel very general: it implies that there are network/connectivity issues, not that the connection is perfectly fine but paypal doesn't like what it's being asked to do. Let's rephrase this just a little to something like "We were unable to establish a connection to PayPal. Please try again in a few minutes". Small change, but implies that it has nothing to do with the user's network conditions.

Pomax (changes requested):

source/js/components/paypal.js
danielfmiranda: Great point Pomax! I have updated the message and re-requested review

Pomax (approved):

PR #1652, opened 2022-04-26
SQS is now first in line, as well as adding a try/except block with error logging

### LINKED ISSUE #1647: Adding try / except and retry for AWS SQS queue that feeds Cinchy
Request asks is to add try / except and retry for SQS queue that feeds Cinchy.  

Since the donate-wagtail app doesn't communicate directly with Cinchy, the Python code is responsible for adding the message to the AWS SQS  queue.  

The code to investigate for possible try / except blocks is likely sqs.py

danielfmiranda:
Closes #1647 
Related PRs/issues #

This PR introduces a try except block, which attempts to use the SQS client's send_message function. 
If this fails for any reason however, it will print an error, wait 5 seconds, and then try again for a maximum of 3 tries.

I also updated the rearranged the order in which two functions fire, **send_to_sqs** and **process_donation_receipt**, as previously if **process_donation_receipt** failed, **send_to_sqs** would not fire at all. 

Unfortunately due to the github issue we cannot create a review app, but I tried to explain what I did / cite documentation in the comments of the PR.

Please let me know if there are any questions!


donate/payments/tasks.py
```diff
@@ -137,8 +137,6 @@ def send_newsletter_subscription_to_basket(data):
 
 
 def send_transaction_to_basket(data):
-    if settings.DONATION_RECEIPT_METHOD == 'DONATE':
-        process_donation_receipt(data)
     send_to_sqs({
```
danielfmiranda, line 140: While looking at this portion of code with Pomax, we found that if the "process_donation_receipt" function returns an exception or errors out, the "send_to_sqs" function is never called. To fix this, I moved the SQS function before the process_donation_receipt function, that way in the event that it fails, the information would still get sent to SQS

donate/payments/sqs.py
```diff
@@ -19,13 +23,25 @@ def sqs_client():
...
+    for n in range(send_message_retries):
+        try:
+            client.send_message(
+                QueueUrl=settings.BASKET_SQS_QUEUE_URL,
+                MessageBody=json.dumps(payload, cls=DjangoJSONEncoder, sort_keys=True),
+            )
+
+            break
+
+        except botocore.exceptions.ClientError as err:
```
danielfmiranda, line 43: Using the [boto3 docs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/error-handling.html), I am listening for any error with the client while trying to use the "send_message" function, if there is one, the logger prints it out and will wait 5 seconds before trying a maximum of 3 times.

donate/payments/sqs.py
```diff
@@ -19,13 +23,25 @@ def sqs_client():
 
 
 def send_to_sqs(payload):
+    send_message_retries = 3
     # If BASKET_SQS_QUEUE_URL is not configured, do nothing (djangorq is logging the payload).
     if settings.BASKET_SQS_QUEUE_URL:
         client = sqs_client()
         if client is None:
+            logger.error("Could not connect to SQS Client.")
```
danielfmiranda, line 31: I added this because I thought it would be a good idea to let the team know if we cant connect to the client for some reason instead of just skipping with no warning

donate/payments/tasks.py
Pomax: If this can throw, in addition to moving it, let's also make sure to try/except the call so that we can log when it throws.

Pomax (changes requested):

donate/payments/tasks.py
danielfmiranda: Hi Pomax! Thanks for the heads up, the try/except is being added in https://github.com/mozilla/donate-wagtail/pull/1654. I will request review once I update the sleep to use sched instead 👍

donate/payments/sqs.py
```diff
@@ -18,14 +24,40 @@ def sqs_client():
         )
 
 
+def attempt_send_to_sqs(client, payload):
```
danielfmiranda, line 27: Similar to my trouble in https://github.com/mozilla/donate-wagtail/pull/1654/files#r861483363, I moved the actual "attempt" to a seperate function, and am sending the items the call needs through **call_args**.

danielfmiranda:
Hi @Pomax thanks for the feedback on this and my other try/except ticket. I have used your feedback from both tickets to update this PR, and have re-requested review. 

Thanks!

donate/payments/sqs.py
```diff
@@ -1,11 +1,17 @@
 from functools import lru_cache
 import json
-
+import logging
+import time
```
Pomax, line 4: just to prevent confusion: `from time import time, sleep` and then a bit further down you  don't need the kind of weird `time.time` (yay python naming =)

donate/payments/sqs.py
```diff
@@ -1,11 +1,17 @@
...
+import sched
 from django.conf import settings
 from django.core.serializers.json import DjangoJSONEncoder
 
+import botocore
 import boto3
 
+logger = logging.getLogger(__name__)
+schedule = sched.scheduler(time.time, time.sleep)
```
Pomax, line 13:
```suggestion
schedule = sched.scheduler(time, sleep)
```

donate/payments/sqs.py
Pomax, line 4:
```suggestion
from time import time, sleep
```

donate/payments/sqs.py
```diff
@@ -18,14 +24,40 @@ def sqs_client():
...
+            schedule.enter(send_data_immediately, schedule_priority, attempt_send_to_sqs, call_args)
+        else:
+            schedule.enter(send_data_delay, schedule_priority, attempt_send_to_sqs, call_args)
+        try:
+            # explicitly block, just in case the implicit behaviour changes in the future
+            schedule.run(blocking=True)
+            break
+        except botocore.exceptions.ClientError as err:
+            if attempt != 2:
+                logger.error(f"Error when sending data to SQS: {err.response['Error']['Message']}")
```
Pomax, line 59: I'm pretty sure this is safe, but `err.response['Error']['Message']` is always tricky: what happens when `Message` doesn't exist? Or one level up, `Error`? It'd be better to make sure these exist before we try to template them into the string.

Pomax (changes requested):
apologies for the delay!

donate/payments/sqs.py
Pomax: @danielfmiranda I'd still recommend making sure this is a safe attribute access

Pomax (changes requested):

donate/payments/sqs.py
```diff
@@ -18,14 +24,40 @@ def sqs_client():
...
+            schedule.enter(send_data_immediately, schedule_priority, attempt_send_to_sqs, call_args)
+        else:
+            schedule.enter(send_data_delay, schedule_priority, attempt_send_to_sqs, call_args)
+        try:
+            # explicitly block, just in case the implicit behaviour changes in the future
+            schedule.run(blocking=True)
+            break
+        except botocore.exceptions.ClientError as err:
+            if attempt != 2:
+                logger.error(f"Error when sending data to SQS: {err}")
```
danielfmiranda, line 59: Hi @Pomax!

While working on the safe attribute access, I decided to see what `err` itself prints as. 

What I found was it actually displays a more detailed message that I think would be of use, in case the donate stack is having trouble connecting to SQS.

For example, here is an error message that was returned when I added a false queueURL to the sendmessage function:
`An error occurred (AWS.SimpleQueueService.NonExistentQueue) when calling the SendMessage operation: The specified queue does not exist for this wsdl version.`

I am now voting for printing the entire error instead of err.response['Error']['Message]' for the following reasons:
- Avoids the need to do a safe attribute check
- Prints where and why the error occured, which would be more helpful to people troubleshooting

Pomax: agreed, that's much better.

Pomax (approved):

PR #1654, opened 2022-04-27
Receipt sending now has try except block for maximum of 3 times

### LINKED ISSUE #1646: Check response messsage and add retry logic for to Acoustic Transact
### Request is to check response message and add retry logic for to Acoustic Transact. 

E.g. add try / except blocks to see what happened to message around: https://github.com/mozilla/donate-wagtail/blob/master/donate/payments/tasks.py#L79-L118

### Requirements:
1. Add try / except logic around Acoustic Transact
2. Write out a log / error message should message fail after a certain number of times.

danielfmiranda:
Closes #1646 

This PR introduces a new try/except block around the function that sends donation receipts. If there is an error connecting to Acoustic as Dan G mentioned has happened in the past, it will log that there was an error, wait 5 seconds, and then try to send the message again for a maximum of 3 times. 

Also, once #1652 gets merged in as well, donator data will be sent to SQS everytime. So even in the event that there is a connection error while sending the receipt, the donators data can be found in SQS to send a receipt later.


donate/utility/acoustic/acoustic.py
```diff
@@ -107,10 +108,23 @@ def __init__(self, client_id, client_secret, refresh_token, server_number):
         super().__init__(client_id, client_secret, refresh_token, server_number)
 
     def _call_xt(self, xml):
+        send_retries = 3
         logger.debug("Request: %s" % xml)
-        response = self.session.post(
-            self.api_xt_endpoint, data=force_bytes(xml), timeout=10,
-        )
+        for n in range(send_retries):
```
danielfmiranda, line 113: Similar to the _call function above, we are now trying the post request that is sending a message, if there is a connection error which is the error that Dan G mentioned he experienced, we are then logging that we could not send the message due to a connection error.

jzinner (changes requested):

donate/utility/acoustic/acoustic.py
```diff
@@ -106,7 +109,8 @@ def __init__(self, client_id, client_secret, refresh_token, server_number):
         self.api_xt_endpoint = self.api_xt_endpoint % server_number
         super().__init__(client_id, client_secret, refresh_token, server_number)
 
-    def _call_xt(self, xml):
+    def _call_xt(self, to, campaign_id, fields, bcc, save_to_db):
+        xml = transact_xml(to, campaign_id, fields, bcc, save_to_db)
```
danielfmiranda, line 113: The scheduler really did not seem to like when I would try to send one "xml" object as a argument even after trying various methods, such as running transact_xml, the function which takes and formats the xml, and then passing it to the scheduler. 

So instead, I am now passing the required fields as arguments and then converting the xml in _call_xt.

danielfmiranda:
Thanks @jzinner for the great feedback! I have implemented everything that was requested and re-requested review 😄 

Hi @Pomax! In this PR I am trying to use the sched module as discussed in #1652, however, I will admit that I was having some trouble with it. I was wondering if you could please take a look at my use here whenever you have a free moment and let me know what you think?

Thank you!

Pomax:
@danielfmiranda note that `.format` is considered obsolete in favour of using f-strings: `method = "preferred", msg = f"this is {method}"`

donate/utility/acoustic/acoustic.py
```diff
@@ -116,7 +120,28 @@ def _call_xt(self, xml):
...
+            send_retries = 3
+
+            for attempt in range(send_retries):
+
+                if attempt == 0:
+                    # If this is the first try, send the email immediately.
+                    schedule.enter(0, 1, self._call_xt, (to, campaign_id, fields, bcc, save_to_db))
+                else:
+                    # If first try is unsuccessful, wait 10 seconds and try again.
+                    schedule.enter(10, 1, self._call_xt, (to, campaign_id, fields, bcc, save_to_db))
```
Pomax, line 132: Can we explain why we're only waiting 10 seconds? Is there an issue where we agreed on that time, or is there some information that future readers can use to understand why we're using this specific value?

On that note, let's remove some duplication, and make sure we don't have "magic constants" in the code:

```python
    def send_mail(self, to, campaign_id, fields=None, bcc=None, save_to_db=False):
        if settings.TESTING:
            return

        send_mail_immediately = 0  # seconds
        send_mail_delay = 10  # seconds
        schedule_priority = 1
        call_args = (to, campaign_id, fields, bcc, save_to_db)

        for attempt in range(send_retries):
            if attempt == 0:
                schedule.enter(send_mail_immediately, schedule_priority, self.attempt_send_mail, call_args)
            else:
                schedule.enter(send_mail_delay, schedule_priority, self.attempt_send_mail, call_args)

            try:
                schedule.run(blocking=True)  # explicitly block, just in case the implicit behaviour changes in the future
                break
            except ConnectionError:
                ...
```

donate/utility/acoustic/acoustic.py
```diff
@@ -116,7 +120,28 @@ def _call_xt(self, xml):
...
+
+                try:
+                    schedule.run()
+                    break
+
+                except ConnectionError:
+                    if attempt != 2:
+                        logger.error("Error connecting to Acoustic to send email receipt. Trying again.")
+                    else:
+                        logger.error("Could not send email receipt. Unable to connect to Acoustic after 3 retries.")
```
Pomax, line 142:
```suggestion
                        logger.error(f"Could not send email receipt. Unable to connect to Acoustic {send_retries} retries.")
```

Pomax (changes requested):

donate/utility/acoustic/acoustic.py
danielfmiranda: Hi @Pomax, no explicit reason for the 10 seconds, I just added it as I felt having the user wait more than 30 seconds might not be the best experience, but will ping @jzinner to see what they think about the time and whether they would like to change it!

donate/utility/acoustic/acoustic.py
danielfmiranda: Follow up issue https://github.com/mozilla/donate-wagtail/issues/1655 created!

danielfmiranda:
Hi @Pomax thank you for the great suggestions, I have added them and re-requested review! Also pinged @jzinner regarding the 10 seconds delay. I will wait to merge until I get confirmation that is the desired time

donate/utility/acoustic/acoustic.py
jzinner: Does this logic in acoustic.py run in real-time when the user clicks submit?  

@danielfmiranda when you test this in development with 10 second retries, are you experiencing a 30 second wait wait for the retries to finish - until the thank you message appears?  

If this runs in real-time and the user has to wait for the retries then 10 seconds is too long and wait 3 seconds between retries.
If this runs in the background and the user does not have to wait for this to finish, then 10 seconds is ok.

danielfmiranda: Hi @jzinner! 

The thank you message appears as soon as the payment goes through on the braintree side. So this would operate in the second option, which is the user receives the thank you message and the message will try to send again in 10 seconds in the background. 

In that case, I will leave it at 10 seconds. Thank you!

donate/utility/acoustic/acoustic.py
```diff
@@ -1,5 +1,6 @@
 import logging
-
+import sched
+import time
```
Pomax, line 3:
```suggestion
from time import time, sleep
```

donate/utility/acoustic/acoustic.py
```diff
@@ -10,6 +11,7 @@
 
 logger = logging.getLogger(__name__)
 XML_HEADER = '<?xml version="1.0" encoding="UTF-8" standalone="yes"?>'
+schedule = sched.scheduler(time.time, time.sleep)
```
Pomax, line 14:
```suggestion
schedule = sched.scheduler(time, sleep)
```

Pomax (changes requested):
basically R+, just a tiny fix =)

danielfmiranda:
Hi @Pomax! I have implemented your feedback and re-requested review, thanks!

Pomax (approved):

PR #1656, opened 2022-05-09
now able to share using ?amount=other

danielfmiranda:
This PR is related to a request from @anilkanji where he would like people to be able to share the donate page with the "other" amount field highlighted. 

Ideally, I would like to open another ticket to improve/clean this up in the future, as this was requested to be ready by 11AM PST on 5/10/22

### DIFF
--- a/donate/core/models.py
+++ b/donate/core/models.py
@@ -118,6 +118,14 @@ class DonationPage(Page):
         """
         return sorted(initial_currency_info['presets'][initial_frequency])[1]
 
+    def default_other_amount(self, initial_currency_info):
+        """
+        If ?amount = other, return value as minimum possible value.
+        We then check in the template if the 'value' is set to the minimum, and if so, highlight the
+        'other amount' field.
+        """
+        return initial_currency_info['minAmount']
+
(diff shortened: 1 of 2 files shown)


donate/core/models.py
```diff
@@ -118,6 +118,14 @@ def default_initial_amount(self, initial_currency_info, initial_frequency):
         """
         return sorted(initial_currency_info['presets'][initial_frequency])[1]
 
+    def default_other_amount(self, initial_currency_info):
```
danielfmiranda, line 121: If the user shares the link with ?amount=other, we set the returned value to the minimum value of whichever currency we are using.

Then in donate_form.html, we are checking if the 'initial_amount', is == to the minimum amount for the currency. If so, highlight the "Other amount" field.

One thing I would like to note is that if the user does not type a custom amount and clicks "submit" the form still preforms as expected. It refreshes the page as the user left it blank.

fessehaye (approved):

tbrlpld (approved):
Thanks @danielfmiranda! Approved.

anilkanji:
Thanks team!

PR #1657, opened 2022-05-20
403 page

danielfmiranda:
Closes #1653 

Screenshot of page:
image

PR #1658, opened 2022-05-23
403 page l10n fixes

TheoChevalier:
Follow-up from https://github.com/mozilla/donate-wagtail/pull/1657

Hi @danielfmiranda, would you mind reviewing this quick localization fix for the 403 page?

Basically, here’s what it does and why, because it’s not intuitive:
- Moving HTML tags outside the string when possible — here we can move the `<p></p>` outside, to avoid any alteration by mistake from translators as well as unnecessary retranslation in case we touch that code in the future.
- Moving the URL to a variable, for the same reasons
- Adding the "trimmed" option to `blocktrans`, to remove leading/trailing spaces that are adding noise to translations

### DIFF
--- a/donate/templates/403.html
+++ b/donate/templates/403.html
@@ -13,8 +13,8 @@
     <div class="layout__container">
         <div class="layout__half-col error__container">
             <p class="error__message">
-                {% blocktrans with url="https://www.donate.mozilla.org" trimmed %}
-                Please disable your VPN and visit <a href="{{ url }}">donate.mozilla.org</a> again.
+                {% blocktrans with url="/" site=request.get_host trimmed %}
+                Please disable your VPN and visit <a href="{{ url }}">{{ site }}</a> again.
                 {% endblocktrans %}
             </p>
             <p class="error__message">
--- a/donate/templates/403.html
+++ b/donate/templates/403.html
@@ -12,10 +12,12 @@
     </div>
     <div class="layout__container">
         <div class="layout__half-col error__container">
-            {% blocktrans %}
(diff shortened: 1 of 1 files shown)


donate/templates/403.html
```diff
@@ -12,10 +12,12 @@ <h1 class="error__heading heading heading--primary">{% trans "Action Required -
...
     <div class="layout__container">
         <div class="layout__half-col error__container">
-            {% blocktrans %}
-              <p class="error__message">Please disable your VPN and visit <a href="https://www.donate.mozilla.org"> donate.mozilla.org </a> again.</p>
-            {% endblocktrans %}
-          <p class="error__message">
+            <p class="error__message">
+                {% blocktrans with url="https://www.donate.mozilla.org" trimmed %}
+                Please disable your VPN and visit <a href="{{ url }}">donate.mozilla.org</a> again.
+                {% endblocktrans %}
```
Pomax, line 18: Just noticed that the URL has `www` in front of it, which won't resolve, so let's slip a fix for that into this PR too... but as a question, if the URL itself a var, would it make sense to make the URL's text equivalent a variable too, so folks can't change that accidentally?
```suggestion
                {% blocktrans with url="https://donate.mozilla.org" site="donate.mozilla.org" trimmed %}
                Please disable your VPN and visit <a href="{{ url }}">{{ site}}</a> again.
                {% endblocktrans %}
```

Pomax (changes requested):

donate/templates/403.html
TheoChevalier: Yup that makes sense, it’s even better :) 
And your second suggestion seem to work pretty well, we can probably go with that!
I’ll update the PR, thanks!

Pomax (approved):

PR #1683, opened 2022-07-27
updated all inline scripts to use the generated nonce value

### LINKED ISSUE #1672: Donate error in https://donate-wagtail.mofostaging.net/
**Describe the bug**

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'https://donate-wagtail.mofostaging.net/en-GB/'
2. See error below **Donate now** 'An error occurred. Please reload the page or try again later.'

**Expected behavior**
To be able to test donations in the staging site.
As stands the error prevents from filling out any card details in the next screen. 

**Screenshots**
Screenshot 2022-06-29 at 09 38 14

Screenshot 2022-06-29 at 09 49 47

**Desktop (please complete the following information):**
 - macOS Monterey
 - Browser: Chrome

### LINKED ISSUE #1682: Update inline scripts to include the new nonce value
After implementing GA4, which included updating `django_csp` to generate a nonce value, it has been brought to our attention that 'unsafe-inline' will be overridden and that we should pass the generated nonce value to any inline scripts in the code base.

danielfmiranda:
Closed #1682

Since we have updated the django_csp package to now add a nonce value to CSP_SCRIPT_SRC, that means we have to pass that nonce value to any inline scripts in the code base or else they wont run. 

I have searched the whole code base to check for any inline scripts, I found a total of two (though both of these scripts are only rendered if someone is doing AB testing) and have updated them to include the nonce, just in case.

### DIFF
--- a/donate/templates/fragments/ga_events.html
+++ b/donate/templates/fragments/ga_events.html
@@ -2,7 +2,7 @@
 
 {% if data_layer %}
   {{ data_layer|json_script:"data_layer" }}
-  <script>
+  <script nonce="{{request.csp_nonce}}">
     const transactionData = JSON.parse(document.querySelector('#data_layer').textContent)[0][0];
     window.dataLayer = window.dataLayer || [];
     window.dataLayer.push(transactionData);
--- a/donate/templates/fragments/ga_events.html
+++ b/donate/templates/fragments/ga_events.html
@@ -2,7 +2,7 @@
 
 {% if data_layer %}
   {{ data_layer|json_script:"data_layer" }}
-  <script nonce="{{request.csp_nonce}}">
+  <script nonce="{{ request.csp_nonce }}">
     const transactionData = JSON.parse(document.querySelector('#data_layer').textContent)[0][0];
     window.dataLayer = window.dataLayer || [];
     window.dataLayer.push(transactionData);
--- a/donate/templates/payment/includes/trigger_ab_testing_thank_you_event.html
+++ b/donate/templates/payment/includes/trigger_ab_testing_thank_you_event.html
@@ -1,4 +1,4 @@
-<script nonce="{{request.csp_nonce}}">
+<script nonce="{{ request.csp_nonce }}">
     document.addEventListener('DOMContentLoaded', function () {
         // Try to send an A/B signal to the wagtail A/B testing framework
         (function tryWagtailABEvent(count) {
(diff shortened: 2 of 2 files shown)


donate/templates/fragments/ga_events.html
```diff
@@ -2,7 +2,7 @@
 
 {% if data_layer %}
   {{ data_layer|json_script:"data_layer" }}
-  <script>
+  <script nonce="{{request.csp_nonce}}">
```
tbrlpld, line 5: **Nitpick**

```suggestion
  <script nonce="{{ request.csp_nonce }}">
```

donate/templates/payment/includes/trigger_ab_testing_thank_you_event.html
```diff
@@ -1,4 +1,4 @@
-<script>
+<script nonce="{{request.csp_nonce}}">
```
tbrlpld, line 1: **Nitpick**

```suggestion
<script nonce="{{ request.csp_nonce }}">
```

tbrlpld (dismissed):
@danielfmiranda When I run with these changes locally, I am still getting the same error. Are there other configuration changes necessary to make this work?

tbrlpld:
@danielfmiranda I am actually trying to search these changes on the donate homepage but I can not find them in the DOM. Where should I find this script?

tbrlpld:
@danielfmiranda I am not firm on this topic and maybe I am missing some context here. Why do we need this change? From my reading of https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_inline_script it seems like "unsafe-inline" or providing a nonce are alternative solutions to allow inline scripts. Are we removing the "unsafe-inline" configuration (which might be a good idea)? I don't quite understand how "overriding" plays into it.

danielfmiranda:
Hi @tbrlpld & @fessehaye! To give some more context:

We used to have 'unsafe-inline' in our list of variables for `CSP_SCRIPT_SRC`, which meant we were able to run inline scripts without any nonce value passed.

However, we have updated the django_csp package settings to now include a nonce value in the list for `CSP_SCRIPT_SRC`.

Since `CSP_SCRIPT_SRC` now includes a nonce value, that means this value must be passed to any inline scripts found in the code base, which is what this PR is doing. 

Apologies for the confusing language in the initial PR description, and please let me know if you have any more questions!

mtdenton:
@tbrlpld @danielfmiranda that's correct, a nonce and unsafe-inline apparently can't coexist in the same CSP directive so this addresses that. It also unfortunately means that we need to be on the lookout for other inline scripts without a nonce, but it seems like Daniel has it covered.

fessehaye (approved):

PR #1686, opened 2022-08-15
Fix Thunderbird donate link in thank you template.

Sancus:
These thank you messages should just link to give.thunderbird.net, donate.mozilla.org with a language code doesn't redirect correctly, and we shouldn't specify language here anyway. Let the user's browser preference take priority.


TheoChevalier (approved):
Thanks @Sancus!

PR #1708, opened 2022-11-29
#9505 #1711 Reduce smallest monthly preset from 10 to 5 for USD, EUR and GPB

tbrlpld:
## Description

This PR changes the smallest presets value for monthly donations from 10 to 5 for USD, EUR and GBP. 

This change would have collided with the current min amount for these currencies. 
It was communicated that the desired path to resolve this conflict is to enforce a different minimum amount for monthly than for single donations. 
This PR also includes that change of logic. 

Further more, it was discovered that some of the upsell donation values are below the minimum monthly donation. 
This was raised and discussed in #1711, and the chosen resolution was to raise all upsell amounts that are below the minimum monthly amount to that minimum. 
The different levels are kept, in case we want to re-estabilish the previous values at some point. 

Related PRs/issues mozilla/foundation.mozilla.org#9505 #1711

## Checklist

_Remove unnecessary checks_

**Tests**
- [x] Is the code I'm adding covered by tests?

**Changes in Models:**
- ~~[ ] Did I update or add new fake data?~~
- ~~[ ] Did I squash my migration?~~
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

**Documentation:**
- ~~[ ] Is my code documented?~~
- [x] Did I update the documentation? -> Yes, I have added some info about the PayPal/Braintree config that is necessary.

### DIFF
--- a/donate/payments/forms.py
+++ b/donate/payments/forms.py
@@ -104,6 +104,7 @@ class PostalCodeMixin():
 class StartCardPaymentForm(MinimumCurrencyAmountMixin, forms.Form):
     amount = forms.DecimalField(label=_('Amount'), min_value=0.01, max_value=MAX_AMOUNT_VALUE, decimal_places=2)
     currency = forms.ChoiceField(choices=constants.CURRENCY_CHOICES)
+    frequency = forms.ChoiceField(choices=constants.FREQUENCY_CHOICES, widget=forms.HiddenInput)
     source_page_id = forms.IntegerField(widget=forms.HiddenInput)
 
     def clean_source_page_id(self):
@@ -166,6 +167,11 @@ class CurrencyForm(forms.Form):
 
 class UpsellForm(MinimumCurrencyAmountMixin, forms.Form):
     currency = forms.ChoiceField(choices=constants.CURRENCY_CHOICES, widget=forms.HiddenInput, disabled=True)
+    frequency = forms.ChoiceField(
+        choices=constants.FREQUENCY_CHOICES,
+        widget=forms.HiddenInput,
+        initial=constants.FREQUENCY_MONTHLY,
+    )
     amount = forms.DecimalField(
         label=_('Amount'), min_value=1, max_value=MAX_AMOUNT_VALUE, decimal_places=2,
         widget=forms.NumberInput(attrs={'step': 'any'})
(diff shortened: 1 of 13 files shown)


donate/payments/tests/test_utils.py
```diff
@@ -42,6 +47,46 @@ def test_get_suggested_monthly_upgrade_small_single_amount(self):
         self.assertIsNone(get_suggested_monthly_upgrade('usd', Decimal(1)))
         self.assertIsNone(get_suggested_monthly_upgrade('brl', Decimal(1)))
 
+    def test_get_min_amount_from_currency_info_single(self):
```
danielfmiranda, line 50: **Kudos** great tests, easy to understand 🙌

donate/payments/forms.py
```diff
@@ -157,11 +170,21 @@ class CurrencyForm(forms.Form):
...
+        widget=forms.HiddenInput,
+        initial=constants.FREQUENCY_MONTHLY,
+        required=False,
+    )
     amount = forms.DecimalField(
         label=_('Amount'), min_value=1, max_value=MAX_AMOUNT_VALUE, decimal_places=2,
         widget=forms.NumberInput(attrs={'step': 'any'})
     )
 
+    def clean_frequency(self):
```
danielfmiranda, line 184: **Question** Hey Tibor, I see that this `clean_frequency` is defined, but do not see it being used anywhere else. Just wanted to double check that this was intended?

danielfmiranda (approved):
Hi @tbrlpld! 

Thank you for taking this work on! I can see that it was quite a bit of work especially with updating the existing logic. 

The code looks good to me, after giving it a look.
I can also confirm that it is working as expected after testing on my local machine.

Approved! 👍 just left a small, random question for my understanding if you have the time

donate/payments/forms.py
tbrlpld: Yea, this is a bit of implicit functionality that comes in by extending the Django `forms.Form`. The `clean_frequency` method is triggered under the hood when `is_valid` is called on the form (so are all other methods that follow the pattern `clean_<fieldname>`). 

There is a whole page in the Django docs dedicated to the form and field validation. I guess that's a pointer that this raises questions quite frequently 😅 
https://docs.djangoproject.com/en/4.1/ref/forms/validation/#form-and-field-validation

tbrlpld:
Thanks for reviewing and merging @danielfmiranda

PR #1715, opened 2022-12-09
updated currency settings to account for differences between thunderbird & foundation

danielfmiranda:
While the foundation donate site needed to update its minimum value setting and default donation amounts to combat fraud, Thunderbird brought to our attention that they would like to keep their minimum value at $2, and their default donation amounts the same as before. 

To resolve this issue, this PR does the following:

- Adds a new currency configuration object to `settings/base.py`. This object is set to the Foundation's updated values, and acts as a default.
- Adds a new currency configuration object to the file `settings/thunderbird.py`, which includes the Thunderbird team's desired minimum and default donation values. This file overrides the base settings whenever the env variable `DJANGO_CONFIGURATION` is set to a Thunderbird instance. 
- Updates the `CURRENCIES` variable in `constants.py`, to pull the correct currency configuration from settings.
- Updates `development.py`, `review_app`, `staging.py`, `production.py`, to account for new `CURRENCIES` object in the settings.

## Checklist

**Tests**
- ~~[ ] Is the code I'm adding covered by tests?~~

**Changes in Models:**
- ~~[ ] Did I update or add new fake data?~~
- ~~[ ] Did I squash my migration?~~
- ~~[ ] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?~~

**Documentation:**
- ~~[ ] Is my code documented?~~
- ~~[ ] Did I update the documentation?~~


donate/settings/development.py
```diff
@@ -41,6 +41,7 @@ def post_setup(cls):
 class ThunderbirdDevelopment(Development, ThunderbirdOverrides, Configuration):
     INSTALLED_APPS = ThunderbirdOverrides.INSTALLED_APPS + Development.INSTALLED_APPS
     FRONTEND = ThunderbirdOverrides.FRONTEND
+    CURRENCIES = ThunderbirdOverrides.CURRENCIES
```
tbrlpld, line 44: Is this line necessary? You have already applied that override in the parent class `ThunderbirdOverrides`. It seems like the child classes should inherit that.

donate/settings/production.py
```diff
@@ -34,6 +34,7 @@ def post_setup(cls):
 class ThunderbirdProduction(Production, ThunderbirdOverrides, Configuration):
     INSTALLED_APPS = ThunderbirdOverrides.INSTALLED_APPS + Production.INSTALLED_APPS
     FRONTEND = ThunderbirdOverrides.FRONTEND
+    CURRENCIES = ThunderbirdOverrides.CURRENCIES
```
tbrlpld, line 37: See above.

donate/settings/review_app.py
```diff
@@ -28,6 +28,7 @@ def post_setup(cls):
 class ThunderbirdReviewApp(ReviewApp, ThunderbirdOverrides, Configuration):
     INSTALLED_APPS = ThunderbirdOverrides.INSTALLED_APPS + ReviewApp.INSTALLED_APPS
     FRONTEND = ThunderbirdOverrides.FRONTEND
+    CURRENCIES = ThunderbirdOverrides.CURRENCIES
```
tbrlpld, line 31: See above

donate/settings/staging.py
```diff
@@ -33,6 +33,7 @@ def post_setup(cls):
 class ThunderbirdStaging(Staging, ThunderbirdOverrides, Configuration):
     INSTALLED_APPS = ThunderbirdOverrides.INSTALLED_APPS + Staging.INSTALLED_APPS
     FRONTEND = ThunderbirdOverrides.FRONTEND
+    CURRENCIES = ThunderbirdOverrides.CURRENCIES
```
tbrlpld, line 36: See above

tbrlpld (approved):
@danielfmiranda Thanks for this quick fix. Just one question regarding the need to repeat the override in all child configurations.

donate/settings/development.py
danielfmiranda: Hi Tibor! It appears that it is:
- When I take this line out, the minimum donation amount is then reverted back to $10
- Also, I can see that the other overrides on the parent class `ThunderbirdOverrides` also have to redefine this line in every file, for example `FRONTEND`.

tbrlpld:
@danielfmiranda I looks like the overrides are necessary, because the inheritance order is off. The parent classes are applied right to left.

PR #1731, opened 2023-01-31
prepend `SVB -` to Thunderbird's beneficiary name on 'ways-to-give'

MelissaAutumn:
Hey there!

We were notified we needed to add this to our "Ways to Give" page for the Euros section. So here it is!

Let me know if I've missed any required steps for changing templates.

Cheers,
Mel


mtdenton:
Hi @MelissaAutumn , getting this in front of a dev. Thanks!

danielfmiranda (approved):
Just tested this out and can confirm that it is working as expected! Approved 👍

PR #1737, opened 2023-03-10
added apple pay domain association view

danielfmiranda:
## Description

Related PRs/issues https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10262

**This PR adds the following:**
- A new environment variable `APPLE_PAY_DOMAIN_ASSOCIATION_KEY`
- A new view, `apple_pay_domain_association_view` which returns the contents of the env variable when the url `/.well-known/apple-developer-merchantid-domain-association` is visited
- Tests to make sure that the view returns the appropriate response with and without a key set

## Checklist

**Tests**
- [x] Is the code I'm adding covered by tests?

## Steps to test:

1. Checkout this branch and run `docker-compose up`
2. Visit http://localhost:8000/.well-known/apple-developer-merchantid-domain-association, the page should return the message `Key not found. Please check environment variables.`
3.  Add a test string for the env variable `APPLE_PAY_DOMAIN_ASSOCIATION_KEY`
4. Restart the app
5. Visit http://localhost:8000/.well-known/apple-developer-merchantid-domain-association again, this time, the page should return whatever string you set for the env variable.
6. If everything is working as expected, testing is complete!


donate/payments/tests/test_views.py
```diff
@@ -967,3 +969,33 @@ def test_get_initial_populates_email_from_session(self):
...
+class TestApplePayDomainAssociationView(TestCase):
+    def setUp(self):
+        self.view_url = '/.well-known/apple-developer-merchantid-domain-association'
+        self.factory = RequestFactory()
+
+    def test_view_returns_key(self):
+        """
+        Make sure the view returns the apple pay key when set
+        """
+        settings.APPLE_PAY_DOMAIN_ASSOCIATION_KEY = "test_apple_pay_key"
```
tbrlpld, line 983: It is recommended to not set settings directly to avoid surprising effects from test order: https://docs.djangoproject.com/en/4.1/topics/testing/tools/#overriding-settings

donate/payments/tests/test_views.py
```diff
@@ -967,3 +969,33 @@ def test_get_initial_populates_email_from_session(self):
...
+        request = self.factory.get(self.view_url)
+
+        response = apple_pay_domain_association_view(request)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content.decode(), "test_apple_pay_key")
```
tbrlpld, line 989: What do you think about using the [`assertContains`](https://docs.djangoproject.com/en/4.1/topics/testing/tools/#django.test.SimpleTestCase.assertContains) method? It can check status and content in one go without the need to `decode` the `content`.

donate/payments/tests/test_views.py
```diff
@@ -967,3 +969,33 @@ def test_get_initial_populates_email_from_session(self):
...
+
+    def test_view_with_no_key_set(self):
+        """
+        Make sure the view returns the appropriate error message when no key is set
+        """
+        settings.APPLE_PAY_DOMAIN_ASSOCIATION_KEY = None
```
tbrlpld, line 995: See above comment regarding settings override.

donate/urls.py
```diff
@@ -15,8 +15,7 @@
 from donate.payments import urls as payments_urls
 from donate.payments.braintree_webhooks import BraintreeWebhookView
 from donate.payments.stripe_webhooks import StripeWebhookView
-from donate.views import EnvVariablesView, ThunderbirdRedirectView, WaysToGiveView
-
+from donate.views import EnvVariablesView, ThunderbirdRedirectView, WaysToGiveView, apple_pay_domain_association_view
 # Patterns not subject to i18n
```
tbrlpld, line 19: I think the empty line here go lost on accident. What do you think about adding it back?

env.default
```diff
@@ -28,6 +28,10 @@ AWS_S3_CUSTOM_DOMAIN=
 AWS_SECRET_ACCESS_KEY=
 AWS_STORAGE_BUCKET_NAME=
 
+
```
tbrlpld, line 31: Looks like there already is an empty line above. What do you think about removing this one?

tbrlpld (approved):
Thanks @danielfmiranda. The implementation is solid. However, I have left a few small questions on the test file.

env.default
danielfmiranda: Great catch Tibor, updated!

donate/payments/tests/test_views.py
danielfmiranda: Great suggestion Tibor! Updated 👍

donate/payments/tests/test_views.py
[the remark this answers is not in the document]
danielfmiranda: Updated! Thanks TIbor

PR #1756, opened 2023-05-31
Added new fields and styling for Campaign pages to allow for CTA-first layouts

### LINKED ISSUE #1755: Rework the campaign page template to allow for the CTA to appear first on the page on mobile
Stakeholders: Audrey Hingle, Kevin Hill

This team is trying to build an appeal page similar to what other organizations use (examples given [here](https://donate.redcross.org.uk/appeal/turkey-syria-earthquake-appeal) and [here](https://www.unicef.org.uk/donate/donate-now-to-protect-children-in-ukraine/))

After going back and forth a bit, seems like we want to adapt `campaign_page.html` for this purpose, mostly looking for the CTA to be above the fold in the appeal. There are a few things to consider here however:

- We can't magically spring the intro header (which is an `h1` and should actually be an `h2`, we'll fix that too) into position above everything.
- The desktop layout should remain the same. We have to be careful changing the order of things using CSS since users utilizing screen readers won't have the same experience as others.
- We don't want to retroactively affect any of the other existing pages by changing this template.

To address the above, we need to make the following adjustments to the model and template:
 - Allow for editing of the hard-coded "Donate now" header that appears on all of these pages (defaulting to "Donate now" for all of the existing pages). Leaving this blank will remove the markup.
 - Create a settings flag to allow page creators to rearrange the layout of these pages so that the CTA will appear at the top when the page collapses on smaller sizes. Once again, this should default to unchecked to allow for existing pages to retain their current layout. This flag will set the order of the markup to include the CTA first. While not ideal, we can manipulate the position of these on tablet sizes and above with `order` CSS rules for the various rows that include the donate form, the "intro text", and the disclaimers without differing the experience for screen reader users too much.
 
 ## Dev Tasks
- [x] Create a `cta_first` boolean setting for the `CampaignPage` class. This should determine the order of the markup in `campaign_page.html` (Checked = cta->intro text->disclaimer, unchecked = intro text ->cta->disclaimer)
- [x] Add a field for `intro_header` in `CampaignPage`, replace the hard-coded "Donate now" text
- [x] Create `order` css rules that only apply for tablet/desktop sizes when `cta_first` is `True`
- [x] Change the intro header `h1` to a `h2`

## Acceptance Criteria
- [x] As a user, when "CTA first" setting is unchecked, the donation form stays at the bottom of the page on mobile and remains on the right side on tablet sizes and above.
- [x] As a user, when "CTA first" setting is checked, the donation form goes to the top of the page on mobile and remains on the right side on tablet sizes and above
- [x] As a user, when "CTA first" setting is checked, the "Intro header" text appears above the donation form on mobile sizes and disappears from below the form.
- [x] As a user, when "Intro header" is filled out, an `h2` appears above the "Intro" rich text on the page
- [x] As a user, when "Intro header" is left blank, no `h2` markup appears above the "Intro" rich text on the page

mtdenton:
Closes #1755

This adds a few things to campaign pages:

- `cta_first` boolean that determines the order of the markup
- When `cta_first` is checked, it retains the old layout style on tablet sizes and above
- Replaces hard-coded "Donate now" text with an editable intro section header
- sections of `campaign_page.html` have now been split out into fragments to increase readability

## Checklist

_Remove unnecessary checks_

**Tests**
- [ ] Is the code I'm adding covered by tests?

**Changes in Models:**
- [ ] Did I update or add new fake data?
- [ ] Did I squash my migration?
- [ ] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

**Documentation:**
- [ ] Is my code documented?
- [ ] Did I update the documentation?


donate/templates/fragments/donate_form_fundraise_up.html
```diff
@@ -1,3 +1,12 @@
+{% load static i18n %}
+
+{% if page.donate_form_header != "" %}
+    <h2 class="heading heading--primary">
+        {% if page.donate_form_header != "" %}
```
mtdenton, line 5: I forgot to remove this

donate/core/models.py
```diff
@@ -241,17 +241,32 @@ class CampaignPage(DonationPage):
...
         FieldPanel('project'),
         ImageChooserPanel('hero_image'),
-        FieldPanel('lead_text'),
+        FieldPanel('donate_form_header'),
+        FieldPanel('intro_header'),
         FieldPanel('intro'),
         InlinePanel('donation_amounts', label='Donation amount overrides'),
     ]
 
+    cta_first = models.BooleanField(
```
mmmavis, line 260: From mockup it looks like we want to always keep CTA on the right on desktop regardless. If I'm not misunderstanding it then let's update the help text so it provides more clarity to users.

mmmavis (changes requested):
Thanks for coming up with a solution with better A11Y support. In addition to the inline comments I left, I also have a question regarding if a new form header is needed. Having two big headers on the same viewport seems to make the page a bit busy. Maybe this is something we should check with the design team to get styling suggestions.

image

image

donate/core/models.py
mtdenton: Yep that's correct, I forgot to change the text here after the `order` adjustments

donate/templates/fragments/donate_form_fundraise_up.html
```diff
@@ -1,3 +1,10 @@
+{% load static i18n %}
+
+{% if page.donate_form_header != "" %}
```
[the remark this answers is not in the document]
mtdenton: Lol gotta love Python, of course that's equivalent.

donate/templates/pages/core/campaign_page.html
```diff
@@ -36,22 +36,33 @@ <h1 class="heading heading--primary hero__heading">{{ page.title }}</h1>
         </div>
     </div>
     <div class="layout__container">
-        <div class="layout__half-col">
+        {% if page.cta_first == True %}
```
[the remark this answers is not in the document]
mtdenton: Yeah I wasn't sure why it wasn't done that way in the first place, although if it's for any reason I'd assume that they wanted to make the fragment contents agnostic of the amount of space it's using in a row/column.

danielfmiranda (commented):
Hi @mtdenton thanks for the tag!

From a backend standpoint, this looks good to me! I have also tested it out on my local machine and can confirm that its working as expected.

I see that Mavis has left some design related comments, so I am going to leave my review as a comment, and will leave the approval to Mavis once the design stuff has been taken care of.


Also, in regards to this comment:

> > Thanks for coming up with a solution with better A11Y support. In addition to the inline comments I left, I also have a question regarding if a new form header is needed. Having two big headers on the same viewport seems to make the page a bit busy. Maybe this is something we should check with the design team to get styling suggestions.
> 
> @mmmavis Happy to pull out `donate_header`, just need to find another solution for getting the `intro_header` above the CTA on mobile that doesn't duplicate that `h2` in the HTML. I'm thinking the best route is to pull it out into its own row, so I'm going to play around with that a bit.


I agree that it would be best to have it on its own row so we only need to render the h2 once.

However, if that ends up being a lot of trouble, the landing page achieves this through the use of two elements and the classes [--mobile](https://github.com/MozillaFoundation/donate-wagtail/blob/24b493ba68bdefc41454d2c06b6ebbff099bcad7/donate/templates/pages/core/landing_page.html#L30) and [--desktop](https://github.com/MozillaFoundation/donate-wagtail/blob/24b493ba68bdefc41454d2c06b6ebbff099bcad7/donate/templates/pages/core/landing_page.html#L39), with CSS hiding them at the appropriate breakpoints [here](https://github.com/MozillaFoundation/donate-wagtail/blob/main/source/sass/base/_layout--fundraiseup.scss)

donate/templates/pages/core/campaign_page.html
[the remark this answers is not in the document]
mtdenton: @mmmavis Despite that, I've carved out some fragments here 👍

mmmavis:
Thank you @mtdenton for making the updates.

One last thing - 

I noticed that when cta_first is off, we see two headers on narrow screens

image

donate/templates/fragments/campaign_page_fundraiseup_form.html
```diff
@@ -0,0 +1,12 @@
+{% load static i18n %}
+
+<div class="layout__half-col {{ order }}">
+    <div class="donate-form">
+        <h2 class="heading heading--primary donate-form__heading-mobile">
+            {% trans page.intro_header context "Header" %}
+        </h2>
+        <div class="donate-form__fundraiseup-form">
```
mmmavis, line 8: Let's replace this with `{% include %}` to donate/templates/fragments/donate_form_fundraise_up.html

mtdenton: done!

